PR #40, opened 2015-08-04
Fix port ranges in PF rule generator

### LINKED ISSUE #38: [PATCH] PF rule generator handles port range incorrectly
```
The PF rule generator doesn't emit port ranges correctly, instead it just emits 
only the low and high ports of the range.

Patch against r245.

As an aside: where is capirca moving to now that code is shutting down?
```

Original issue reported on code.google.com by `jethrobe...@gmail.com` on 26 Apr 2015 at 10:25

Attachments:
- [capirca.diff](https://storage.googleapis.com/google-code-attachments/capirca/issue-38/comment-0/capirca.diff)

jethrogb:
The PF rule generator used to emit port ranges incorrectly, instead emitting
only the low and high ports of the range.

cc @mavam

 Fixes #38.

### DIFF
--- a/lib/packetfilter.py
+++ b/lib/packetfilter.py
@@ -240,8 +240,10 @@ class Term(aclgenerator.Term):
   def _GeneratePortStatement(self, ports):
     port_list = []
     for port_tuple in ports:
-      for port in port_tuple:
-        port_list.append(str(port))
+      if port_tuple[0]==port_tuple[1]:
+        port_list.append(str(port_tuple[0]))
+      else:
+        port_list.append(str(port_tuple[0])+':'+str(port_tuple[1]))
     return '{ %s }' % ' '.join(list(set(port_list)))
 
 


vklimovs:
Hey,
Thank you for your interest! This has been fixed internally and pushed to open source version in https://github.com/google/capirca/commit/e07f760f3e07d51688bf2ab2685514419072384c.

lib/packetfilter.py
```diff
@@ -240,8 +240,10 @@ def _GenerateAddrStatement(self, addrs, exclude_addrs):
   def _GeneratePortStatement(self, ports):
     port_list = []
     for port_tuple in ports:
-      for port in port_tuple:
-        port_list.append(str(port))
+      if port_tuple[0]==port_tuple[1]:
+        port_list.append(str(port_tuple[0]))
+      else:
+        port_list.append(str(port_tuple[0])+':'+str(port_tuple[1]))
```
mavam: Nit: looking at the rest of the source, it seems the coding style uses spaces around binary operators. For consistency, would you mind adapting the patch format accordingly?

PR #48, opened 2015-12-30
Added cisco object-group support.  fixes #36

### LINKED ISSUE #36: Cisco object-group ACL not generating as expected
```
What steps will reproduce the problem?

1. Use the included NETWORK.net & SERVICES.svc definition files included & the 
attached policy file, test_cisco_object_group.pol.

2. Run aclgen.py on this policy file:

./aclgen.py -d def -p policies/test_cisco_object_group.pol -o output

3. Review the output file.

What is the expected output? What do you see instead?

I've attached the generated output file, test_cisco_object_group.acl.  It 
creates an object group for GOOGLE_DNS:

object-group ip address GOOGLE_DNS
 8.8.4.4 255.255.255.255
 8.8.8.8 255.255.255.255
exit

But when actually writing the ACL, it uses the IP addresses for the addgroup 
instead of the object group name:

 permit 17 addrgroup 0.0.0.0/0  addrgroup 8.8.4.4/32 portgroup 53-53
 permit 17 addrgroup 0.0.0.0/0  addrgroup 8.8.8.8/32 portgroup 53-53
 permit 17 addrgroup 0.0.0.0/0  addrgroup 2001:4860:4860::8844/128 portgroup 53-53
 permit 17 addrgroup 0.0.0.0/0  addrgroup 2001:4860:4860::8888/128 portgroup 53-53

It also uses a 0.0.0.0/0 address group which isn't a defined object group (I 
would expect it to just use the keyword any).

What version of the product are you using? On what operating system?

SVN revision 259 on OS X 10.6.8 (Python 2.7.1).
```

Original issue reported on code.google.com by `david.co...@gmail.com` on 30 Oct 2014 at 9:23

Attachments:
* [test_cisco_object_group.pol](https://storage.googleapis.com/google-code-attachments/capirca/issue-36/comment-0/test_cisco_object_group.pol)
* [test_cisco_object_group.acl](https://storage.googleapis.com/google-code-attachments/capirca/issue-36/comment-0/test_cisco_object_group.acl)

jbutler42:
Also, cleans up Cisco header comment sequence and placements when using multiple headers for the same target.

Also, switched to using protocol names instead of protocol integers, I am not sure if that is unsupported on some Cisco's (may need to revert this part if so), but our Cisco devices convert the protocol integer to protocol name which makes post verification painful as the version produced by the device is translated and the version produced by Capirca is (was) integer, raising false errors on our post process validation.

Note, we need to get some tests written and shipped with this package.

### DIFF
--- a/aclgen.py
+++ b/aclgen.py
@@ -117,14 +117,15 @@ def revision_tag_handler(fname, text):
   return '\n'.join(new_text)
 
 
-def get_policy_obj(source_file, definitions_obj, optimize, shade_check):
+def get_policy_obj(source_file, definitions_obj, optimize, shade_check,
+                  platform=None):
   """Memoized call to parse policy by file name.
 
   Returns parsed policy object.
   """
 
   return policy.CacheParseFile(source_file, definitions_obj, optimize,
-                               shade_check=shade_check)
+                               shade_check=shade_check, platform=platform)
 
 
 def render_filters(source_file, definitions_obj, shade_check, exp_info):
@@ -173,7 +174,8 @@ def render_filters(source_file, definitions_obj, shade_check, exp_info):
       optimized = this_platform['optimized']
       # Copy Policy Obj.
       pol = copy.deepcopy(get_policy_obj(source_file, definitions_obj,
-                                         optimized, shade_check))
+                                         optimized, shade_check,
+                                         target_platform))
       renderer = this_platform['renderer']
       # Render.
       fw = renderer(pol, exp_info)
--- a/lib/cisco.py
+++ b/lib/cisco.py
@@ -200,49 +200,47 @@ class ObjectGroup(object):
   def __str__(self):
     ret_str = ['\n']
     addresses = {}
-    ports = {}
+    service_items = {}
 
     for term in self.terms:
-      # I don't have an easy way get the token name used in the pol file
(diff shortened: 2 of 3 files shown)


vklimovs:
This is already implemented in internal version, unfortunately. We plan on completely migrating over to open source Capirca in first quarter of 2016 for internal use too (and getting rid of internal version), it's best if you see how it looks then and suggest changes then. 

It was our unfortunate decision to keep internal version for so long.

vklimovs:
Oh and most important part - internal version has tests for all generators.

jbutler42:
Nice! Looking forward to that.  Since Q1 is nearly here, I will just wait for your pull from your internal version before I make any further pull requests.  Definitely interested in seeing how you got the object-group thing working, as my approach was to add a nasty global var to the policy module, giving it the hint that it is processing for a cisco target.  I am sure there are alternative approaches that are far more elegant.

vklimovs:
Sorry for the delay - still very much aiming at merging all internal stuff in this Q (Q2).

jejenone:
I'm very much interested in this fix too.

ankenyr:
I am going to close this pull request. The internal migration was completed and the issue should be resolved. Please let us know if this is untrue.

PR #52, opened 2016-03-15
Included a Palo Alto Firewall target

nexusapoorvacus:


vklimovs:
This is very cool, thank you for doing that. Please add tests (https://github.com/google/capirca/tree/master/tests). We have tests for all other generators internally and we aim to merge them all here - but I'd like all new code from contributors to have tests.

ankenyr:
The current version of capirca now has a lot of tests that can be used as examples. Do you intend to implement some tests for this new generator?

nexusapoorvacus:
Yes, I will include the tests for this generator in the next few weeks. Thank you for letting me know.

ankenyr:
Hi @nexusapoorvacus have you had a chance to update the tests?
There is also a new way to document and validate supported tokens per platform. I would suggest looking at #96.

Also it looks like you still need to sign the CLA. There are instructions above on how to do that. Please do it so I can accept your code.

nexusapoorvacus:
I haven't had a chance to update the tests. I do not have PAF software I can test on any longer. I will try and get a hold of the software, but if anyone else would like to write tests, please feel free.

nexusapoorvacus:
I signed it!

lib/paloaltofw.py
```diff
@@ -0,0 +1,534 @@
...
+        rules.append(self.INDENT * 10 + "<member>" + tz + "</member>")
+      rules.append(self.INDENT * 9 + "</to>")
+
+      rules.append(self.INDENT * 9 + "<from>")
+      for fz in options["from_zone"]:
+        rules.append(self.INDENT * 10 + "<member>" + fz + "</member>")
+      rules.append(self.INDENT * 9 + "</from>")
+
+      rules.append(self.INDENT * 9 + "<source>")
+      if options["source"] == []:
```
ankenyr: You should use an implicit evaluation
please see https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations
and from https://www.python.org/dev/peps/pep-0008/
For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

Yes: if not seq:
     if seq:

No: if len(seq):
    if not len(seq):

lib/paloaltofw.py
```diff
@@ -0,0 +1,534 @@
...
+        rules.append(self.INDENT * 10 + "<member>any</member>")
+      else:
+        for s in options["source"]:
+          rules.append(self.INDENT * 10 + "<member>" + s + "</member>")
+      rules.append(self.INDENT * 9 + "</source>")
+
+      rules.append(self.INDENT * 9 + "<destination>")
+      if options["destination"] == []:
```
ankenyr: Same as above, implicit vs explicit

lib/paloaltofw.py
```diff
@@ -0,0 +1,534 @@
...
+        rules.append(self.INDENT * 10 + "<member>any</member>")
+      else:
+        for d in options["destination"]:
+          rules.append(self.INDENT * 10 + "<member>" + d + "</member>")
```
ankenyr: Line to long, 80 character limit

lib/paloaltofw.py
```diff
@@ -0,0 +1,534 @@
...
+      rules.append(self.INDENT * 9 + "</destination>")
+
+      rules.append(self.INDENT * 9 + "<service>")
+      if options["service"] == []:
```
ankenyr: Same, implicit vs explicit.

lib/paloaltofw.py
```diff
@@ -0,0 +1,534 @@
...
+    rules.append(self.INDENT * 5 + "</rulebase>")
+
+    end = []
+    end.append(self.INDENT * 4 + "</entry>")
+    end.append(self.INDENT * 3 + "</vsys>")
+    end.append(self.INDENT * 2 + "</entry>")
+    end.append(self.INDENT * 1 + "</devices>")
+    end.append("</config>")
+
+    return '\n'.join(initial) + '\n\n' +'\n'.join(service) + "\n\n" + '\n'.join(rules) + '\n' + '\n'.join(end)
```
ankenyr: Line too long, 80 character limit

policies/sample_paloalto.pol
```diff
@@ -0,0 +1,68 @@
...
+################
+
+# NOTE: makes sure zone names correspond to those specified in your firewall setup
+
+header {
+  target:: paloalto from-zone Trust-L3 to-zone Trust-L3
+}
+
+term allow-ping-firewall{
+  destination-address:: FIREWALL_EXTERNAL
```
ankenyr: FIREWALL_EXTERNAL doesn't exist in our .net files. Can you please fix this? Either add something into multitarget_sample.pol (I did this to test some of your code) or fix this file to use things already in the .net files.

tools/update-pa-firewall.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/python
```
ankenyr: Is there a reason you wish to add this file? I am not doubting that it may be useful to you but will it be useful to others in their own environment? I don't use palo alto firewalls so I am not exactly sure. There is also the fact that by accepting this file we are on-boarding the responsibilities around the file. We could say that things under the tools directory are unsupported but I would rather not do that either.

Let me know the reasoning behind this addition.

XioNoX:
@ankenyr 
How can I contribute to that PR?
I have those patches working on the current master. In addition I added:
* Address book/set support
* PAN application support (> layer 4)
* Raises errors if policies/services names are too long (61 char for services, 31 for policies)

ankenyr:
Hey @XioNoX,
This all depends on how well the commits went. If you can do the work to separate it out into a couple PRs I would be happy to review and accept them in short order. I would want an initial PR that is this initial add with the comments I mentioned fixed. Another adding tests. Then follow ups for each of the features above.
If that seems reasonable then I will look forward to your PR's. If not lets discuss a better strategy.

XioNoX:
@ankenyr 
I have everything in 1 paloaltofw.py file, so rolling back and separating it in several commits would be quite a pain.
I did however addressed the remarks you made on the current file, as well as running it through autopep8 (via atom beautify).
I haven't had time to look at the tests, but would like someone to review my current work before investing more time.

ankenyr:
I don't want to accept the code until tests are made for it. I am happy to do a review of a PR without tests just to make sure things look sane. After that if you can get some tests out we will be able to accept it. Does that sound fair? If so go ahead and start a new PR and we can get this started.

XioNoX:
Yeah, that's what I was looking for! Thanks.
https://github.com/google/capirca/pull/113 opened!

ankenyr:
I am going to go ahead and close this and follow up with #113. Thanks for the initial work @nexusapoorvacus. Thanks for picking things up @XioNoX

PR #55, opened 2016-04-06
Add aclgen smoketests

jzohrab:
This is a preparatory PR that should let us add characterization tests to the existing code.  It adds some simple smoke tests to `aclgen.py`, generating to the existing project folders.  The output is _not_ checked, these are smoke tests only.

The tests can be run from the project root dir with `$ python -m unittest discover`.

If/when this PR is accepted, I can add some very simple characterization tests, which would consist of running acl with the existing committed data, storing the output, and using that as the basis for test assertions, which would allow for work on the code without breaking anything.

**Summary of code changes:**

* `main()` parameterized to `main(array_of_args)` so unit tests can pass in different arg lists.
* Added `test` folder to keep this distinct from `tests` folder.  I don't know if the existing `tests` actually work, and didn't want to expand the scope of this PR.
* passing the FLAGS.output_directory explicitly to routings that need it, rather than using a global FLAGS variable.  FLAGS could have been left as global, but for the tests to work it would need to be marked as 'global' in a few places which felt ugly.  I can undo that change if preferred.

**Notes:**

* `test/test_aclgen.py` captures the stdout/sterr output of the test runs.  This assumes that the folder only contain information which has been committed as part of the base project.  If people are making independent changes _in this source code_, it's going to be hard to track.  Another option would be to add an array of items that we think should be checked (that is, the "official" source code).
* I haven't added anything to the README about how to run the tests (`$ python -m unittest discover`), but can do so

### DIFF
--- a/aclgen.py
+++ b/aclgen.py
@@ -28,6 +28,7 @@ import datetime
 from optparse import OptionParser
 import os
 import logging
+import sys
 
 # compiler imports
 from lib import naming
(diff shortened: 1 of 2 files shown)


jzohrab:
Thanks for merging, @vklimovs!

PR #56, opened 2016-04-09
Add characterization tests

jzohrab:
This PR adds some characterization tests to ACL generation, so we can be sure that changes to the existing code don't break any of the generation routines.  The test data files are the same as the current sample def and policy files in the project root.

This is as large-looking PR, but it is broken down into a few self-contained commits with messages explaining their intent.

Some preparatory work was needed for these tests to be added, those are the 5 commits leading up to the "Add characterization tests" commit:

* 924df32 Improve test error output.
* 65d1862 Ignore warnings printed to stderr during smoke test.
* 9e228e3 Refactor: handle revision tags before output.
* eb4306b Allow suppression of revision info in generated files.  _note: this adds a new cli flag, necessary to simplify the tests_
* 52a3d21 Fix aclgen directory output for tests.

The "Add char. tests" commit is large, as it adds the files for the characterizations.

* 1dcdc80 Add characterization tests.

This last commit is to ensure that the directory structure in the generated files matches the input.  See the commit notes for information, also ref the notes in commit 52a3d21.  There are sanity check test cases for this routine in particular:

* 7f8b506 Generate output dir structure mirroring the input.

I'm happy to discuss this over email or via skype/whatever, as you may be concerned about pulling in a large-looking PR.  I can split it up into smaller PRs as well - I believe each of these commits is self-contained and passes all tests at each point in time, though I may have broken something during the rebase.  At any event, this full PR works and passes all tests (`python -m unittest discover`).

Some possible future work:

- the aclgen.py methods are starting to need many params.  We could pass all of those around in a single FLAGs argument, rather than passing each individually.  I think that would still be clear.  I didn't make that change as I felt that this change was already quite big, and I wanted to have some baseline tests in place before going any further.

Regards,
Jeff


jzohrab:
There are some elements in the Capirca grammar that are not covered by the sample policies.  e.g., the "qos" element generates a "forwarding-class" in the sample_juniper_loopback.jcl.  @vklimovs, should I add that test coverage to this PR, or wait until you've reviewed this one?  I don't want to force more things into your work/review queue.

vklimovs:
Internal version has pretty extensive tests for all generators. It's rather painful for us to have two versions - so we are going to deprecate (remove) internal one and only develop open source version. But there is no real need to implement extensive generator tests independently.

jzohrab:
OK, these were simple to add given the existing examples.  Thanks.

jzohrab:
Per discussion with Robert @ankenyr today, I'm going to work on my development branch and get it up-to-speed with the latest changes in master.  I'll then resubmit this PR in some other shape. jz

PR #67, opened 2016-06-04
LARGE PR: all of my development branch changes

jzohrab:
This is a very large pull request, it will likely be easiest to discuss this in a conference call.

I created this branch (merge_master) from my development branch, https://github.com/jzohrab/capirca/commits/development, and then merged google Capirca master into it.  I first fixed the merge conflicts, and then did individual commits to get the tests running and passing.  List of these commits: https://github.com/jzohrab/capirca/commits/merge_master.  If needed, I could squash all of these fixes down into one _giant_ commit, but felt it would be better to have them separate at first.

All the unit and integration tests in this branch currently run and pass with "python -m unittest discover".  This includes the tests I added, as well as the old tests and the new ones added to master.  "Ran 497 tests in 52.177s   ...  OK", which is pretty nice.  Thanks for adding all the tests.

Capirca/master was updated as I did this work, and so I merged that work, as at commit 840616e, in here as well.

# Changes

This PR merges almost all of the changes I've made in my development branch (see an overview of these changes at the list of closed pull requests into my dev branch: https://github.com/jzohrab/capirca/pulls?q=is%3Apr+is%3Aclosed).

## Biggest changes

* parsing of the Capirca data is separated from the model
* added a yaml policy parser, and a "policy_format" flag so users can choose which format to parse.
* added a YamlNaming object, similar to the existing Naming object for consistency
* added characterization tests for the existing code, and test coverage for the above

Note that the yaml format does _not_ have complete feature parity with the capirca format, this is documented in doc/policy_format_yml.md, and perhaps should be iterated on.  We're using this format at the moment for an internal POC project.

I backed out some of my changes (eg the introduction of the AclGen object in the series of commits in my PR 2, commit d120259) since I chose to use the new aclgen.py code in Capirca master, following discussion with Rob (ankenyr), so even though those commits are in the rev history the content is overwritten with the merge at commit b48fba9.

## Smaller changes

* ~~added cisco nexus support~~ _UPDATE_: The code is here but is not activated due to issues raised by @supertylerc.  I'll fix the code and either append here or open a new PR
* cleaned up some logging, for people wishing to use the library modules as a third-party library (which we're doing in a project at the moment)
* pip requirements
* fixed the new tests from Google to use the policyparser object
* added symlinks for test discovery (see "dummy symlinks" note below)
* some documentation clean-up

# Notes

## Choosing and documenting one test method

I added tests when the public release of Capirca had no automated tests.  These are in "test" and modules are named "test_..." to allow for test discovery.

The newly released master code has some tests under "tests".  We should combine these, but I've left them separate for now as I didn't want to introduce yet more churn, and I didn't know what was a good choice.

I also recommend that we change the code so that all of them can be run from a single command-line call - perhaps we could rename the modules in the "tests" directory from "xxx_test.py" to "test_xxx.py" to facilitate test discovery.

## Dummy symlinks

Symlinks added in commit 03f1e1712.

'python -m unittest discover' requires that all tests be named 'test_xxx.py'. Renaming all of the 'tests' files to this naming convention would be extra thrash,
and perhaps would not be the right decision for the team/project.  This hack lets us run all of the tests in one shot using test discovery, at least until the CI direction is set.

### Error not raised on `naming.NoDefinitionsError` in aclgen:

Note that files are still generated even if there is a bad def’s directory:

```
  except naming.NoDefinitionsError:
    logging.fatal('bad definitions directory: %s', FLAGS.definitions_directory)
```

and the error is not rethrown.

### Potential implementation issue with aclgen.DescendRecursively

I'm not sure what the current DescendRecursively function is aiming to do.  It appears to ignore nested directories, eg the below:

```
pol
  - some_pol.pol
  nested
    - some_nested_dir.pol  << IGNORED
```

I removed that from the characterization data test case.  We should add some tests around that to ensure it's doing exactly what you expect.

---

Thanks, and sorry for the giant wall of text.  Let me know when you're free to discuss, it would be great to get these into master.


supertylerc:
I had an issue with this on when using the `cisconx` target.  The output had the IP protocol numbers instead of the protocol name (`permit 17` instead of `permit udp`, for example), and it was also using the IOS mask wildcard instead of slash notation (`0.0.0.0` instead of `/32`, for example).  This was seen when using the YAML parser.  I haven't tried the non-YAML parser with this branch.

Everything works as expected on @jzohrab's `development` branch.

jzohrab:
Thanks @supertylerc - I made some changes to the class to reduce the impact on the cisco tests that google added, and must have broken the Nexus support.  It was working in the dev branch, so I'll disable Nexus support in this branch, work to restore the output, and will likely open a new PR.

jzohrab:
@supertylerc and PR reviewers: rather than revert the commit which adds Nexus support to the main line - I dislike reverts in the commit history - I rebased and got rid of it entirely.  I'll work on that in a separate branch and ensure it passes the tests in the development branch, which I believe are correct, and then either append to this branch or open a new PR.

jzohrab:
Opened a new PR for Nexus support which builds on this PR.

aclgen.py
```diff
@@ -46,10 +46,14 @@
 from lib import packetfilter
 from lib import pcap
 from lib import policy
+from lib import policyparser
 from lib import speedway
 from lib import srxlo
 from lib import windows_advfirewall
```
ankenyr: Please remove the line here

aclgen.py
```diff
@@ -97,7 +101,11 @@
     'exp_info',
     2,
     'Print a info message when a term is set to expire in that many weeks.')
-
+flags.DEFINE_enum(
+    'policy_format',
+    'capirca',
+    ['capirca', 'yml'],
+    'policy file format')
```
ankenyr: Please leave two lines between top level entries

aclgen.py
```diff
@@ -181,14 +189,23 @@ def RenderFile(input_file, output_directory, definitions,
     logging.warn('bad file: \n%s', e)
     raise
 
+  filename, file_extension = os.path.splitext(input_file)
+  format_dict = {
+    '.yml': yamlpolicyparser,
```
ankenyr: add 2 spaces on this and the next line.

aclgen.py
```diff
@@ -181,14 +189,23 @@ def RenderFile(input_file, output_directory, definitions,
...
+    '.pol': policyparser
+  }
+  if file_extension not in format_dict:
+    raise Error('bad policy file extension ' + file_extension)
```
ankenyr: Please create a specific Error class and add this to the docstring for RenderFile

aclgen.py
```diff
@@ -368,11 +385,12 @@ def FilesUpdated(file_name, file_string):
   return False
 
 
-def DescendRecursively(input_dirname, output_dirname, definitions, depth=1):
+def DescendRecursively(input_dirname, policy_file_extension, output_dirname, definitions, depth=1):
```
ankenyr: This line is to long, we use 80 characters as the max line length

aclgen.py
```diff
@@ -434,18 +453,35 @@ def WriteFiles(write_files):
...
+                'policy_file: %s\nrendered_acl_directory: %s\n'
+                'policy_format: %s',
                 str(sys.argv[0]),
                 int(FLAGS.optimize),
                 str(FLAGS.base_directory),
                 str(FLAGS.policy_file),
-                str(FLAGS.output_directory))
+                str(FLAGS.output_directory),
+                str(FLAGS.policy_format)
+  )
```
ankenyr: Wrong indentation level, please add 13 spaces

aclgen.py
```diff
@@ -434,18 +453,35 @@ def WriteFiles(write_files):
...
+
+  format_dict = {
+    'yml': ('.yml', yamlnaming.YamlNaming),
```
ankenyr: Add 2 spaces here and the next line

jzohrab:
@ankenyr - Is there a style guide/tool you're using for these checks?  I can check these things if so.

lib/__init__.py
```diff
@@ -0,0 +1,19 @@
+__version__ = '1.1.0'
```
ankenyr: Maybe I can learn something here. If we just remove __init__ does this not just include everything in lib? This was changed by @finfinack in a previous ef1a05917f8445d82c77002cc416227e4d992d84 and things seem to work well enough without it.

ankenyr:
Ohh jeez, each of those are individual. Awesome. Sorry for the spam.
I sent you an email mentioning I was going to do a first pass of lint errors, and I was hoping it would be all as one.

Our style guide can be found here https://google.github.io/styleguide/pyguide.html

lib/__init__.py
jzohrab: My project was using the lib items directly, and not calling the top-level script, so I _believe_ we need to have init.py for the module to be usable.  I can test it out and verify.

ankenyr:
Looks line the following is a chromium plugin for pylint
https://chromium.googlesource.com/chromiumos/chromite/+/master/cli/cros/lint.py
I believe this may get you to where you need for a linter to check against our style guide.

jzohrab:
np about the spam, I'm happy to get all of it for this, and thanks for going through everything!

Thanks for the link to the style guide.  I'll set aside some time to work through the pylint details.  When I work through it, I can add some docs as well (in a separate PR).

aclgen.py
jzohrab: This policyparser is how Capirca .pol policies actually get parsed.  If I remove this.

aclgen.py
jzohrab: Added at 536b865 in #69

lib/__init__.py
finfinack: It may work without it but adding the __init__.py files tells python what the structure of the project is - what can be imported from where. If you don't import cross folders it probably works without. I mainly added it after reading up on it and consider it cleaner to add them everywhere or remove them altogether but please ignore me if anyone here knows more about __init__.py...

jzohrab:
Opened PR #69 for review and discussion, instead of working on this branch.

lib/__init__.py
jzohrab: It's necessary, it ensures that the dir is treated as a package.

* https://docs.python.org/3/tutorial/modules.html
* http://stackoverflow.com/questions/448271/what-is-init-py-for

In particular, if we want to export the renderers as a separate library (which I'm currently doing for my project, and I hope to extend for use with Napalm/Ansible/other tools), we need it.

jzohrab:
After discussion with @ankenyr, I've broken this giant PR up into several smaller PRs which are now open.

I'll open a separate issue for one of the points above.  I've addressed some of the others (i.e., choosing a test file naming convention, following Google's lead).

PR #68, opened 2016-06-04
Add Nexus support

jzohrab:
Add Nexus support.

Builds on PR #67.  Only the tip of this branch (commit ccfef745) is new.

This contains updated test cases for Nexus which look correct.  Leaving this as a separate PR so that I can address any issues in it separately from PR 67.


jzohrab:
Replaced by PR #74.

PR #71, opened 2016-06-10
Add test coverage

jzohrab:
Added characterization tests to ensure that behaviour is preserved during upcoming changes and refactoring.


aclgen.py
```diff
@@ -434,7 +434,11 @@ def WriteFiles(write_files):
     output.flush()
 
 
-def main(_):
+def main(args):
```
ankenyr: There is no need for this section. Tests would pass if you did the FLAGS.Reset() in the setup or tear down phase in the tests.

tests/integration/aclgen_test.py
```diff
@@ -0,0 +1,181 @@
...
+import aclgen
+from lib import policy
+from lib import naming
+
+
+class Test_AclGen_ensure_demo_files_work(unittest.TestCase):
+  """Ensure Capirca demo runs successfully out-of-the-box."""
+
+  def setUp(self):
+    self.iobuff = StringIO()
```
ankenyr: Above do a similar import as in aclgen for gflags as flags.
Then between the class definition and imports add "FLAGS = flags.FLAGS" and somewhere in the setup method add FLAGS.Reset() and things should work.

tests/integration/aclgen_test.py
```diff
@@ -0,0 +1,181 @@
...
+    logger.level = logging.DEBUG
+    self.s = logging.StreamHandler(self.iobuff)
+    self.s.level = logging.DEBUG
+    logger.addHandler(self.s)
+
+    # Capirca only writes if files have changed, so write out to a new
+    # temp dir for each test run.
+    self.output_dir = tempfile.mkdtemp()
+
+    curr_dir = os.path.dirname(os.path.abspath(inspect.getfile(inspect.currentframe())))
```
ankenyr: Are you sure this is the cleanest way to do things? The variable says curr_dir which I assume to be the current directory.
os.getcwd() provides the current working directory. If you just made self.root_dir = os.getcwd() this looks to pass.

tests/integration/aclgen_test.py
```diff
@@ -0,0 +1,181 @@
...
+    self.root_dir = os.path.realpath(os.path.join(curr_dir, '..', '..'))
+    self.policies_dir = os.path.join(self.root_dir, 'policies')
+    self.defs_dir = os.path.join(self.root_dir, 'def')
+
+  def tearDown(self):
+    # print self.iobuff.getvalue().split("\n")
```
ankenyr: can this be removed?

tests/integration/aclgen_test.py
```diff
@@ -0,0 +1,181 @@
...
+    # Remove the directory after the test
+    shutil.rmtree(self.output_dir)
+
+  def test_smoke_test_generates_successfully(self):
+    args = [
+      'program',  # Dummy value for gflags, which expects the program name to be the first entry.
+      '--base_directory={0}'.format(self.policies_dir),
```
ankenyr: using .format here is a bit overkill. String format is slower than a simple modulo format. We can leave this how it is and save it for a cleanup afterwards.

tests/integration/aclgen_test.py
```diff
@@ -0,0 +1,181 @@
...
+    args = [
+      'program',  # Dummy value for gflags, which expects the program name to be the first entry.
+      '--base_directory={0}'.format(pol_dir),
+      '--definitions_directory=/some_missing_dir/',
+      '--output_directory={0}'.format(self.output_dir)
+    ]
+
+    aclgen.main(args)
+
+    # NOTE that the code still continues work, even if a bad directory
```
ankenyr: Care to explain this further?

tests/lib/naming_test.py
```diff
@@ -169,5 +171,123 @@ def testParseServiceFile(self):
     filedefs._ParseFile(data, 'services')
     self.assertEqual(filedefs.GetService('HTTP'), ['80/tcp'])
 
+
+class Naming_Characterization_Tests(unittest.TestCase):
```
ankenyr: A lot of these look to be exact duplicates of the unittests. Does naming.py really need characterization tests?

aclgen.py
jzohrab: True.  FLAGS is global state, I prefer code to not to have to worry about global state.  We can remove this and do what you suggest if preferred.

tests/integration/aclgen_test.py
jzohrab: I had the exact same question from members of another project I contributed to :-).  This is the way I've been used to doing it, but it could be outdated practice.  I learned of the __file__ variable and it could simplify this.

tests/integration/aclgen_test.py
jzohrab: Which part, the teardown, or the comments?  Yes to the latter, and probably to the former, would have to try it out.

tests/integration/aclgen_test.py
jzohrab: Interesting, ok thanks.

tests/integration/aclgen_test.py
jzohrab: This is related to something I documented in a closed PR: "Error not raised on naming.NoDefinitionsError in aclgen".
Files are still generated even if there is a bad def’s directory, the error is logged as fatal but swallowed.

```
  except naming.NoDefinitionsError:
    logging.fatal('bad definitions directory: %s', FLAGS.definitions_directory)
```

I didn't think that this was the intended behaviour, and added this test for discussion.  At any rate, this test documents what's currently happening.

tests/lib/naming_test.py
jzohrab: Hm, not sure!  When I added these tests, there weren't any unit tests.  I was creating a yaml version of this class, and needed to ensure that the behaviour was consistent.  I couldn't figure out what the behaviour was without running it, and figured that I may as well commit this exploration in the form of tests for the overall project health.

Also, the naming.py class looked like it could use some refactoring, so these tests would make that safe.  "If it ain't broke, don't fix it" notwithstanding, I think there may be some style issues too ... really wasn't sure.

tests/integration/aclgen_test.py
ankenyr: This really should be simplified. os.getcwd() tells me exactly what it is doing. The way you have it using inspect was very weird. I don't think I have ever had to use inspect ever. I actually dropped into a debugger to make sure I understood what you were doing.

tests/integration/aclgen_test.py
ankenyr: The comment, the removing of the tmp directory is fine. The directory will be cleaned up regardless by the OS but it is a good habit to keep your environment clean.

tests/integration/aclgen_test.py
ankenyr: Just as a fyi http://tomatohater.com/2013/08/30/python-performance-string-formatting/
That is a quick blurb showing the difference in speed.
https://waymoot.org/home/python_string/ this is an article talking about string concatenation which I believe helped shape some of the decisions on how we join strings in capirca.

In future cases if you are trying to format a complex string then format is probably the prefered method but for a single variable there is no need for this.

jzohrab: Thanks @ankenyr!

tests/lib/naming_test.py
ankenyr: Sure, I am all for new tests that test different functionality but you have a test_GetIpParents and then there is another test that already exists called testGetIpParents. These look to be exactly the same. Duplicate tests don't really help.

The following look to be duplicates of existing tests
test_GetIpParents
test_GetServiceParent
test_GetNetParent
test_GetService
test_GetService_raises_if_unknown
test_GetServiceByProto
test_GetNet
test_GetNet_raises_if_unknown
test_parseServiceList

These both looks to be unneeded as they only call another test.
test_GetNetAddr
test_GetNetAddr_raises_if_unknown

The other ones should be moved into the normal testing class. In your comment you admit that these are not really characterization tests.

jzohrab: Got it, np.  I was more concerned with getting the code into master, figuring that any duplication could be removed later.

They sort of are characterization tests, in that I wanted to explore and preserve the existing behaviour for a different implementation.  But that's semantic, and these can probably be removed.

tests/integration/aclgen_test.py
jzohrab: fyi, here's the StackOverflow post I first referred to for this: http://stackoverflow.com/questions/50499/how-do-i-get-the-path-and-name-of-the-file-that-is-currently-executing/

`getcwd` could give incorrect directories, potentially.  I believe I could run the tests from project root or different locations.  I've been advised to use `__file__` by others, and have done so in other projects, and so will use that.

aclgen.py
jzohrab: Changed, will repush.

tests/integration/aclgen_test.py
jzohrab: Done, tests passing.

tests/integration/aclgen_test.py
jzohrab: Changed.

tests/integration/aclgen_test.py
jzohrab: Done, thanks.

tests/lib/naming_test.py
jzohrab: @ankenyr, do you think there's any value for users to see what method calls with the existing data would return?  Maybe I could recast this somehow so it's some form of documentation.

Re those tests calling other tests - I have literally no idea what I was thinking with this.  Thanks, removing.

ankenyr: Sorry I am not exactly sure what you mean? Are you talking about making tests that use NETWORK.def or some file? I feel that naming.py is pretty straight forward in its use and does not really require this. If you would like to rework documentation that is fine but I don't believe our documentation is lacking in this regard.

https://github.com/google/capirca/wiki/Naming-definitions
https://github.com/google/capirca/wiki/Naming-library
https://github.com/google/capirca/wiki/Policy-format

Between those three pages it decribes the naming format, the methods, and how they are used in policies.

jzohrab: NP, I'd added this when there were no tests.  I was wondering if you could think of any use for the effort here, if there was anything useful to salvage.  If not, we should scrap it, I'm ok to do that too.  Scrap it?

ankenyr: The duplicate tests should be removed.

jzohrab: @ankenyr - I've removed these duplicate tests, thanks.

smmadge:
Is there any progress on the NXOS integration/merge yet?

PR #72, opened 2016-06-10
Split parsing from policy

jzohrab:
Builds on PR #71.

policy.py is split into policyparser.py and policy.py.

Rationale for this PR: I wanted to introduce a new parser for a project, and couldn't determine where the separation was between the Policy object (the domain model) and the lex/yacc parser.  The lex/yacc parser defines many checks which other parsers/data formats would need, so it made sense to try to cleanly separate the parsing from the domain model, and strengthen the domain model as needed.

A stronger domain model has the following potential benefits:
- easier to load the model from different data sources, and ensure it's still correct
- simpler unit testing (can directly interact with the model, rather than loading up the lex/yacc parser with strings)
- can expose the base domain model for other clients, such as NAPALM or Ansible

This change reflects my own opinion and approach of software design (viz, separation of the domain model from how it's loaded, a domain-driven design approach).  The team/community may prefer to not go down this route.  Regardless, I think this work is still useful in more clearly demarcating the role of domain object vs parser, and even if we choose to recombine the policyparser.py and policy.py files into a single file, the work is still beneficial.


ankenyr:
Going to close this. There is pending work in salt to actually integrate this to work with NAPALM. Since this functionality is only really in use for that framework I would like to keep that specific code there instead of integrating it here where it would increase the complexity of our own codebase.

PR #73, opened 2016-06-10
Add yml parser

jzohrab:
Builds on PR #72.
- adds a yaml policy parser, and a "policy_format" flag so users can choose which format to parse.
- adds a YamlNaming object, similar to the existing Naming object for consistency
- adds characterization tests for this parser

Note that the yaml format does not have complete feature parity with the capirca format, this is documented in doc/policy_format_yml.md, and perhaps should be iterated on. We're using this format at the moment for an internal POC project.


ankenyr:
See #72

PR #75, opened 2016-06-10
Add NullHandler logger

jzohrab:
Builds on PR #74 

lib should log to NullHandler by default.

Capirca classes could potentially be exposed as a library for use via Ansible or similar.  Libraries should use NullHandler for loggers by default.  ref http://docs.python-guide.org/en/latest/writing/logging/.


ankenyr:
See #72

PR #79, opened 2016-06-15
cgrep features and tests

ryantierney513:
Summary:
1. move cgrep out of tools directory and into base directory per Google
2. add several new cgrep features
3. add new PPP class to port.py for service object comparisons (used by new cgrep feature)
4. add 28 cgrep test cases

Test Plan:
test cases within cgrep_test.py and peer review:

```
>cgrep_test.py
............................
Ran 28 tests in 0.068s

OK
```


tools/cgrep.py
```diff
@@ -1,93 +0,0 @@
-# Copyright 2011 Google Inc.
```
ankenyr: Why the move? You should be able to set things up to import from the tools directory correct?

cgrep.py
```diff
@@ -0,0 +1,443 @@
...
+#   $ cgrep.py -p 22 tcp
+#
+
+import argparse
+import pprint
+from lib import naming
+from lib import nacaddr
+
+
+__author__ = ("watson@google.com (Tony Watson)", "jhrbek@fb.com (Joe Hrbek)"
```
ankenyr: Do you really want to be added as an author here? I believe authors are no longer required per Googles code standards and I have been contemplating removing all __author__ tags everywhere. Git maintains a much better history and can show exactly how much someone has contributed. I am not saying you do not deserve credit on this file. You have changed a very significant portion of this file. I am wondering what your opinions are on authors tags in general since this is the first time I have seen someone add themselves to the authors.

cgrep.py
```diff
@@ -0,0 +1,443 @@
...
+  )
+
+  parser.add_argument("-d", "--def", dest="defs",
+                      help="Network Definitions directory location. \n",
+                      default="./def")
+
+  # -i and -t can be used together, but not with any other option.
+  ip_group = parser.add_argument_group()
+  # take 1 or more IPs
+  ip_group.add_argument("-i", "--ip", dest="ip", nargs='+', type=is_valid_ip,
```
ankenyr: You have some inconsistant quote use here and throughout the file. Please fix these.

cgrep.py
```diff
@@ -0,0 +1,443 @@
...
+  :returns None:
+  """
+  options = parser.parse_args()
+  db = naming.Naming(options.defs)
+  p = pprint.PrettyPrinter(indent=1, depth=4, width=1).pprint
+
+  # if -i and any other option:
+  if options.ip and any([options.gmp, options.cmp, options.obj, options.svc,
+                         options.port]):
+    print("You can only use -i with -t or by itself")
```
ankenyr: Have you thought about using logging instead of simple print statements? Also the parens are superfluous, I won't force you to change them though. Give logging some thought though.

cgrep.py
```diff
@@ -0,0 +1,443 @@
...
+
+  # if -p
+  elif options.port:
+    port, protocol, result = get_services(options, db)
+    print("%s/%s:" % (port, protocol))
+    p(result)
+
+  # if nothing is passed
+  elif not any((options.cmp, options.ip, options.token, options.obj,
+               options.svc, options.port)):
```
ankenyr: Please add a single space to line up the return.

cgrep.py
```diff
@@ -0,0 +1,443 @@
...
+    parser.print_help()
+  print('')
+
+
+def check_encapsulated(obj_type, first_obj, second_obj, db):
+  """ Checks if a network/service object is entirely contained within another.
```
ankenyr: I see this method raises an error, usually this is documented in the doc strings also.

cgrep.py
```diff
@@ -0,0 +1,443 @@
...
+
+  :ip str: the IP we're looking for the parents of.
+  :db naming.Naming(): network and service definitions
+  :returns list: a list of all groups that include the IP, in the format:
+           [("Group", ["networks", "matched"]), (etc)]
+  """
+  results = []
+  rval = db.GetIpParents(ip)
+  for v in rval:
+    nested = db._GetParents(v, db.networks)
```
ankenyr: Any reason you are using _GetParents instead of the GetNetParents?
PEP8 defined single underscore prefixed names as being for internal use.

tests/cgrep_test.py
```diff
@@ -0,0 +1,535 @@
...
+"""Unit tests for cgrep.
+
+   Order doesn't matter for the purposes of these tests, so many
+   actual and expected results are sorted/sets to prevent issues relating to the
+   order in which items are returned."""
+
+import unittest
+import sys
+import argparse
+sys.path.append('../')
```
ankenyr: This should not be needed. About to log off and commenting from my laptop so I don't have my standard environment but try the following from the root capirca directory

python -m unittest discover -s . -p '*_test.py'

tests/cgrep_test.py
```diff
@@ -0,0 +1,535 @@
...
+from lib import naming
+from lib import nacaddr
+from cgrep import get_ip_parents
```
ankenyr: Why did you decide to do it like this? To shorten the length of lines? Do you find it easier to read?

ankenyr:
Need to dig in more but I made a few comments to get you started on some stuff.

cgrep.py
usernameisnottaken: I'm in favor of removing author tags everywhere, if people really want to know who did what you should look at version control.

ryantierney513: I agree that Git maintains better history and am fine with removing this. In general I try to follow what I see in the file, but if things have since changed, that's fine by me and the others!

cgrep.py
ryantierney513: Internally, we had developed a slightly different version of _GetParents (with a slightly different name) and called it directly instead of also changing GetNetParents to call it. When ripping that out, I reverted it back to just _GetParents out of habit.
Will update it.

cgrep.py
ryantierney513: We have, but opt'd to follow what we saw used in this file already and keep using print. In other parts of capirca, I think it makes sense to use logging so we can control logging verbosity and consistent format, but in the case of cgrep, since we're just sending text to the user via stdout, I felt it was OK to keep using print.

tests/cgrep_test.py
ryantierney513: Will do

tests/cgrep_test.py
ryantierney513: I initially made the tests with just one import, and then it grew to 2, and so on, and I just kept repeating over time as more functions were added and tests needed to be made. I think it's better that these all be imported in a single statement though.

tools/cgrep.py
ryantierney513: That idea came from vklimovs's email:
"cgrep is fairly small, so let's treat it as first party script (perhaps move out of tools folder)."

That was fine for us, so I tossed it into capirca/. If you'd like it moved back into tools/, I don't think it's a problem either way.

cgrep.py
ryantierney513: I've gone through the various files and tried to standardize on single quote.

cgrep.py
ankenyr: Understood, I figured there was a reason. You may wish to add Facebook Inc to the AUTHORS file located https://github.com/google/capirca/blob/master/AUTHORS at some point. That file is required to exist per Max whom I believe you have talked to before. It should be reserved for organization names and not individuals.

tools/cgrep.py
ankenyr: I will let this be then. I may move it back or into a new folder during some future reorganization.

cgrep.py
ankenyr: Understood, we can leave it as it is now. I feel a good change in the future would be to move it completely over to using logging.
Filed https://github.com/google/capirca/issues/87 to track this. If you wish to take care of it feel free or else I will get to it at some point.

cgrep.py
```diff
@@ -0,0 +1,440 @@
...
+  networks = []
+  ip = nacaddr.IP(ip)
+  # loop through all the networks in the net_group
+  for net in get_nets([net_group], db)[0][1]:
+    # find the highest prefix length for the networks that contain the IP
+    if ip in net:
+      networks.append(str(net))
+      if net.prefixlen > highest_prefix_length:
+        highest_prefix_length = net.prefixlen
+  return (highest_prefix_length, networks)
```
ankenyr: I believe that the parens here are not required.

cgrep.py
```diff
@@ -0,0 +1,440 @@
...
+#   $ cgrep.py -s FOO
+#
+#   To find which service tokens contain port '22' and protocol 'tcp' use
+#   $ cgrep.py -p 22 tcp
+#
+
+import argparse
+import pprint
+from lib import naming
+from lib import nacaddr
```
ankenyr: Please move this import before naming

cgrep.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+def is_valid_ip(arg):
+  """ Validates a value to be an IP or not.
```
ankenyr: Please use the doc string format that is in use in other files.

tests/cgrep_test.py
```diff
@@ -0,0 +1,533 @@
...
+   Order doesn't matter for the purposes of these tests, so many
+   actual and expected results are sorted/sets to prevent issues relating to the
+   order in which items are returned."""
+
+import unittest
+import argparse
+from lib import naming
+from lib import nacaddr
+from cgrep import (
+  check_encapsulated,
```
ankenyr: You want to add two spaces here and the following lines

tests/cgrep_test.py
```diff
@@ -0,0 +1,533 @@
...
+  def test_one_ipv6_fail(self):
+    expected_results = [('LINKLOCAL', ['FE80::/10'])]
+    ip = '2001:db8::1'
+    results = get_ip_parents(ip, self.db)
+    self.assertNotEquals(results, expected_results)
+
+  # 8.8.8.8 is in GOOGLE_PUBLIC_DNS_ANYCAST which is inside GOOGLE_DNS
+  def test_one_ip_nested(self):
+    expected_results = sorted((('GOOGLE_DNS', ['8.8.8.8/32']),
+                              ('GOOGLE_DNS -> GOOGLE_PUBLIC_DNS_ANYCAST',
```
ankenyr: Missing a space on this line and 257

tests/cgrep_test.py
```diff
@@ -0,0 +1,533 @@
...
+    ip = '2001:4860:4860::8844/128'
+    results = sorted(get_ip_parents(ip, self.db))
+    self.assertEquals(results, expected_results)
+
+  # 1.0.0.1 is inside INTERNAL_SERVER, which is inside INTERNAL_SERVERS, which
+  # is inside SERVERS
+  def test_one_ip_multi_nested(self):
+    expected_results = sorted((('INTERNAL_SERVERS -> INTERNAL_SERVER',
+                               ['1.0.0.1/32']),
+                              ('SERVERS -> INTERNAL_SERVER', ['1.0.0.1/32']),
```
ankenyr: Same problem here as above.

tests/cgrep_test.py
```diff
@@ -0,0 +1,533 @@
...
+    options = Namespace(
+        ip=('69.171.239.12',),
+        token=('GOOGLE_DNS')
+    )
+    results = compare_ip_token(options, self.db)
+    self.assertEquals(results, expected_results)
+
+  # 2a03:2880:fffe:c:face:b00c:0:35 is not in GOOGLE_DNS
+  def test_ipv6_in_token_fail(self):
+    expected_results = r'2a03:2880:fffe:c:face:b00c:0:35 is _not_ ' \
```
ankenyr: Please try not to use \ to join strings.
https://google.github.io/styleguide/pyguide.html?showone=Line_length#Line_length

ankenyr:
Sorry for any delays. Once these comments are addressed I will do a pull. This looks great and thank you for doing the work. Cgrep has needed love for a long time.

ankenyr:
Hi Ryan, tried to perform the pull but when I do the import it seems to want to remove all the tests at the new location. I believe you need to rebase. Let me know when you do this and I should be able to import quickly.

ryantierney513:
OK, I've rebased and pushed again. Let me know how it looks now. Thanks.

PR #81, opened 2016-06-16
Filecomment lexxer support

usernameisnottaken:
Summary:
It can be useful to sometimes embed arbitrary comments at the very top of
the file, rather than a few lines in via a header comment. This is supported
with a "file-comment" header attribute that on supported platforms will include
that content at the very top of the generated policy.

Test Plan:
Added new token to aclgenerator unit test - unittest did not
crash.

### DIFF
--- a/lib/cisco.py
+++ b/lib/cisco.py
@@ -829,6 +829,9 @@ class Cisco(aclgenerator.ACLGenerator):
           for line in comment.split('\n'):
             target.append('remark %s' % line)
 
+        for line in header.filecomment:
+          target_header.append("! %s !" % line.replace('"', ""))
+
         # now add the terms
         for term in terms:
           term_str = str(term)
--- a/lib/juniper.py
+++ b/lib/juniper.py
@@ -776,6 +776,7 @@ class Juniper(aclgenerator.ACLGenerator):
 
   def _TranslatePolicy(self, pol, exp_info):
     self.juniper_policies = []
+    self.file_comment = []
     current_date = datetime.datetime.utcnow().date()
     exp_info_date = current_date + datetime.timedelta(weeks=exp_info)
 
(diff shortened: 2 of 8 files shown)


ankenyr:
I don't really see the point. What would you want at the top of the file that couldn't be included in the comments already present? The concept makes it sound like there should only be one of these per file but you can easily have multiple headers in a single file. Can you explain how this is intended to solve and why the comments as they exist now do not provide the required support?

Also your tests do not change any of the expected results. I would suspect those need to change.

sneakywombat:
ankenyr - it's a difference between comments that appear in the middle of rendered output vs the top.  When multiple comments are encountered, they are appended in order and rendered as such.  We use this functionality for internal tools to consume, it's basically a spot we can smash in meta data about the rendered output.

The tests should reflect this behavior, so agreed there. :)  Is it sufficient if usernameisnottaken updates the tests to reflect this?  I think the difference in behavior will be more clear this way.

ankenyr:
Can you share what metadata you would want to have at the top of the file which cannot be used as a top level comment at each header like would be done if you did this in the header comments field? Or at least what function this metadata serves? Is it to something like the version number to track what is running on a firewall?

ankenyr:
Just following up on this

ankenyr:
ping

sneakywombat:
Ryan is going to pick this up. FYI.

ankenyr:
Shall I close this PR expecting a new one?

sneakywombat:
@ryantierney513 has volunteered to pick up where @usernameisnottaken left off.  Ryan?

ankenyr:
Ping @ryantierney513 any chance this can be resolved?

ankenyr:
Same here as #85

PR #82, opened 2016-06-16
Fix nsxv unittest

usernameisnottaken:
Summary:
The nsxv unittest was referencing a text address family
attribute that seems to be Cisco-specific, which is not implemented by
this platform.

Test Plan:
ran all test*.py modules in tests/ directory and saw this
passed.


ankenyr:
Rebase this and we should be gtg, thanks for catching this!

ankenyr:
Hey @usernameisnottaken please resolve the conflict and I will take this ASAP

usernameisnottaken:
Sorry for the delays - getting caught up on all the feedback now!

ankenyr:
Sorry about that, was trying to get travis working and I didn't realize it would run things just yet. It should be turned off at this point.

ankenyr:
Believe this is failing because .travis.yaml was not created before you started. Local tests look to work. Merging.

PR #83, opened 2016-06-16
add platform token validation code and sub token validation code.

sneakywombat:
Summary:
This code adds token and sub token validation for all platforms in google/capirca on github.
I made a good effort to inspect the code and capture all of the token use cases, but there may be some
omissions. Essentially what this will do is define a strict set of tokens that each platform implements.
These tokens are then inspected in __init__ of AclGenerator before the the respective platform _TranslatePolicy.
This will prevent unexpected renderings of policy, as sometimes platforms can silently drop tokens the user provides.
It also enforces actions in a potentially disagreeable way, but does so explicitly. Users will know straight away if a token or sub token they are using will not work as expected.

Test Plan:
All of the unit tests under tests/ were executed and passed successfully.

### DIFF
--- a/lib/nsxv.py
+++ b/lib/nsxv.py
@@ -211,7 +211,7 @@ class Term(aclgenerator.Term):
                                                       af=self.filter_type))
           return ''
         if not source_addr:
-          source_addr.append(source_address)
+          source_addr.extend(source_address)
         else:
           source_addr = source_address
 
@@ -230,11 +230,7 @@ class Term(aclgenerator.Term):
                                                       direction='destination',
                                                       af=self.filter_type))
           return ''
-        destination_addr.append(destination_address)
-    # creating single list out of lists of lists
-    source_addresses = [val for sublist in source_addr for val in sublist]
-    destination_addresses = [val for sublist in destination_addr
-                             for val in sublist]
+        destination_addr.extend(destination_address)
 
     # ports
     source_port = None
@@ -250,9 +246,9 @@ class Term(aclgenerator.Term):
       log = 'true'
 
     sources = ''
-    if source_addresses:
+    if source_addr:
       sources = '<sources excluded="false">'
-      for saddr in source_addresses:
+      for saddr in source_addr:
 
         # inet4
         if type(saddr) is nacaddr.IPv4:
@@ -278,9 +274,9 @@ class Term(aclgenerator.Term):
       sources = '%s%s' %(sources, '</sources>')
 
     destinations = ''
-    if destination_addresses:
+    if destination_addr:
       destinations = '<destinations excluded="false">'
-      for daddr in destination_addresses:
+      for daddr in destination_addr:
         # inet4
         if type(daddr) is nacaddr.IPv4:
           if daddr.numhosts > 1:


sneakywombat:
not to poke the bear on this one, but any comments here?

ankenyr:
Hey @sneakywombat, I took a brief look at this and will need to see how it interacts with our internal processes. I will try and make some time tomorrow to do that. This may end up conflicting with #85. Do you or @usernameisnottaken have a preference on how to resolve this? It seems to me that it would be easier to merge the squid platform and then have you update this PR

lib/aruba.py
```diff
@@ -94,6 +89,13 @@ def _TranslatePolicy(self, pol, exp_info):
       if 'ipv6' in filter_options:
         ip_ver = 6
       new_terms = []
+      for t in terms:
```
ankenyr: I am guessing you are looping through twice to make it only log once. Have you given any thought to creating a logging filter? This could be useful in other areas of the code.

lib/aclgenerator.py
```diff
@@ -280,32 +254,100 @@ def __init__(self, pol, exp_info):
...
 
+  def _buildTokens(self):
+    """provide a default for supported tokens and sub tokens.
+
+    Returns:
+      tuple
+    """
+    # Set of supported keywords for a given platform.  Values should be in
+    # undercase form, eg, icmp_type (not icmp-type)
+    supported_tokens = {'action',
```
ankenyr: Do you have a preference to use {} to define sets? Through out the project we have used it as like 
`set([foo, bar, baz])`

I kind of like it in that form since the curly braces are shared with dicts.

lib/policy.py
```diff
@@ -2151,7 +2151,7 @@ def ParseFile(filename, definitions=None, optimize=True, base_dir='',
 
 
 def ParsePolicy(data, definitions=None, optimize=True, base_dir='',
-                shade_check=False):
+                shade_check=False, filename=''):
```
ankenyr: I see you are adding this to ParsePolicy but should you did not add this to the call from ParseFile. Would it not be a good idea to add it there?

ankenyr:
Made a couple more comments. When you get a moment can you resolve the conflicts present? From a glance it seems it mostly has to do with the test file location.

lib/aruba.py
sneakywombat: i just added some comment support into this platform, it previously didn't support it. :(  iirc, another platform followed this pattern and I was just trying to stick to it.

lib/aclgenerator.py
sneakywombat: I tend to prefer literals over builtins, but the real reason i used it was from this:

`In [1]: from dis import dis

In [2]: dis('{1}')
  1           0 LOAD_CONST               0 (1)
              3 BUILD_SET                1
              6 RETURN_VALUE

In [3]: dis('set([1])')
  1           0 LOAD_NAME                0 (set)
              3 LOAD_CONST               0 (1)
              6 BUILD_LIST               1
              9 CALL_FUNCTION            1 (1 positional, 0 keyword pair)
             12 RETURN_VALUE
`

using the set() call just makes more calls under the hood. The other reason is that it makes set comps easier as well:

`In [4]: bar=[1,2,3,3,4,4,5]

In [5]: {foo for foo in bar}
Out[5]: {1, 2, 3, 4, 5}`

lib/aruba.py
ankenyr: Yea, some platforms do it out of necessity, and I am sure there are some that do it when it is not needed. This isn't a big deal, more an observation and sparking the idea of a future optimization that we could think about.

lib/aclgenerator.py
ankenyr: Well TIL!
Lets leave this then. A possible cleanup/optimization work in the future would be to profile things how they are now and then move everything over to using {} and see if we can make things a bit faster.

lib/policy.py
sneakywombat: Ooo, good call. I didn't think of that.  will do.

sneakywombat:
@ankenyr will fix up the filename in parsefile, let me know what you want me to do with the set changes.  The filter recommendation I agree with as well, we have started doing that in our internal code as well, but it isn't finished yet.  It'd probably be better to do that in another PR though as it would be pretty big.

ankenyr:
SGTM

As for the sets. Feel free to leave it or move it to be consistent with the rest of the project. I do want consistency but I plan on doing some testing to see if we replace everything to {} if we could make things noticeably faster. If so I will change everything over to that format. If there is no noticeable change I will update yours to be in line with the set() standard.

I personally think set() is more readable but I don't feel so strongly that a performance boost couldn't convince me out of that.
Sound good? If so I will make a issue to track my testing.

sneakywombat:
Yeah, the set usage i have here is mostly at the front end of the load, so i'm unsure on how it will influence the speed.  Any speed up is good though right? :)  changes on sunday or monday inbound.

ankenyr:
Hey @sneakywombat did you have a chance to fix this yet?

sneakywombat:
No :( I'm terrible. I'll try to get to it tomorrow.

lib/policy.py
```diff
@@ -2146,12 +2146,12 @@ def ParseFile(filename, definitions=None, optimize=True, base_dir='',
   """
   data = _ReadFile(filename)
   p = ParsePolicy(data, definitions, optimize, base_dir=base_dir,
-                  shade_check=shade_check)
+                  shade_check=shade_check, filename=filename)
```
sneakywombat: boom. done.

lib/policy.py
```diff
@@ -2129,7 +2129,7 @@ def _Preprocess(data, max_depth=5, base_dir=''):
 
 
 def ParseFile(filename, definitions=None, optimize=True, base_dir='',
-              shade_check=False):
+              shade_check=False, filename=''):
```
sneakywombat: and here.

sneakywombat:
@ankenyr i finally updated this! sorry it took so long.  let me know if you want any further changes.

lib/policy.py
sneakywombat: finished.  thanks @ankenyr for the catch here.

ankenyr:
Looks like things are still conflicted? The marker above has not changed.

sneakywombat:
the hashes are different, did i miss something else? https://github.com/google/capirca/pull/83/commits/39cf5b3f9bc3130960fd654f5120770ffebfa6ae

ankenyr:
I *think* this may be because it is not synced to the latest? Looking at git log I see 840616e915b6a3845d4bf3013f9c5dc6a8988318 as the last log from before your changes. That is not the latest bits. I actually have another thing to push out today which changes cisco format slightly. Which will need to be integrated but I don't think it will conflict with you.

I see two choices going forward if my assumption are correct.
1) You rebase to the latest. I don't think there are any conflicts but github does this warning when there are any file changes upstream that were changed.
2) I can patch your stuff as it is internal and push it out. I cannot guarantee you will get credit on github if you care about the stats. I believe our pipeline now supports the ORIGINAL_AUTHOR in the commits but I have not tested this.

I would prefer 1 since it should be easy for you and makes my life easier.

sneakywombat:
Hmm. I did rebase, but will redo.

ankenyr:
Also, just a comment about your commits. They are both the same. Not sure if this is an artifact of mechanisms you use to push things out. I know we have some weird things considering we use https://github.com/google/MOE. If possible having commits that are descriptive of the change being made are helpful. How could I easily tell from the logs what the difference between 39cf5b3f9bc3130960fd654f5120770ffebfa6ae and dc4e1e5a08ee81e79896624738c3cc6f99a3ee29

sneakywombat:
Might be because I did a commit --amend?

ankenyr:
Could be? I have used those to scrub our commits before of things like co-workers emails or internal only links. I have not seen it do this but considering how complex git can get I wouldn't be surprised if you could end up in that state somehow. Either way just a nit picky comment. I appreciate all the work you have put into this.

sneakywombat:
Looks like i maybe got it this time?  This workflow is so much different from what i'm used to.  so sorry.  Let me know if this looks better.

ankenyr:
Hey @sneakywombat still seem to be have a problem. After you did the rebase did you do update the PR?

It seems to think that cgrep doesn't exist.

https://github.com/sneakywombat/capirca
vs
https://github.com/sneakywombat/capirca/tree/39cf5b3f9bc3130960fd654f5120770ffebfa6ae

Indeed it seems that in the PR itself it doesn't exist. This ends up causing some weird issues with me translating things.

sneakywombat:
gah.  ok.  I may just abort the whole thing and start over with a new pr.
I'm not sure what I did on my end to cause this, but i'm sure it was
something I did, or didn't do in git land :(

ankenyr:
Mmmm lets not give up just yet

https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

This seems to show some instructions. I think you just need to perform the last step doing a 'git push -f'

ankenyr:
If you do decide to make a new PR though let me know so I can push out the cisco changes I have waiting. It will be one less thing for me to juggle and shouldn't conflict any of your changes.

ankenyr:
@sneakywombat let me know what you decide. If you could though leave this one open for a while longer. One of the developers thinks I discovered a bug with this so they would like a chance to investigate.

sneakywombat:
Uh oh. Bugs. :) haven't had time to unbreak my local git branch yet.

PR #84, opened 2016-06-17
Support strings that start with integers

usernameisnottaken:
Summary:
It's valid for a policer policy to start with numbers, but the
capirca policy will treat anything starting with an integer as an int type. This adjusts the
regular expression to not include cases where a number is followed by
word character.

Test Plan: Unit tests


ankenyr:
Same here as #85

PR #93, opened 2016-07-22
fix cgrep bug detailed in issue#92

ryantierney513:
Summary: fix cgrep bug detailed in issue#92
cgrep get_services function doesn't respect protocol when comparing against a range

Test Plan: new unit tests, locally tested, internally peer reviewed


ryantierney513:
All the tests pass except for the two NSX tests which are known to fail (and I think fixed in another PR?).

PR #96, opened 2016-08-31
add supported tokens and supported subtokens (retry of pr83)

sneakywombat:
(retry of https://github.com/google/capirca/pull/83) 

This code adds token and sub token validation for all platforms.  I made a good effort to inspect the code and capture all of the token use cases, but there may be some omissions. Essentially what this will do is define a strict set of tokens that each platform implements, instead of relying on the platform to maybe provide what it supports.  Making it mandatory that a platform defines what it implements will make testing easier and also create a better experience for users writing rules against a platform.  They will know immediately that the term they wrote will actually render as they intended it to.

These tokens are inspected in __init__ of AclGenerator before the the respective platform _TranslatePolicy.  This will prevent unexpected renders of policy, as sometimes platforms can silently drop tokens the user provides. It also enforces actions, in a potentially disagreeable way, but does so explicitly.  Users will know straight away if a token or sub token they are using will not work as expected, and explicitly as they have defined it in the term itself.


lib/aclgenerator.py
```diff
@@ -271,48 +249,118 @@ class ACLGenerator(object):
...
     self.policy = pol
-
+    all_err = []
     for header, terms in pol.filters:
       if self._PLATFORM in header.platforms:
         # Verify valid keywords
         # error on unsupported optional keywords that could result
         # in dangerous or unexpected results
         for term in terms:
+          if term.platform:
```
sneakywombat: this was removed in error.  I just popped it back in, and put it on top of the err = [] line so as to prevent that list creation for terms we don't need it on, eg they would be skipped from these checks.

sneakywombat:
@ankenyr, how do you think this looks?

ankenyr:
tested this internal, there are some things missing, collecting everything together to show what would break because they are seen as supported by the platform. Will update soon. Will also take a closer look afterwards on each platform to find other missing ones.

ankenyr:
keywords (destination_prefix) for target cisco in policy
 keywords (established) for target pcap in policy
 keywords (first-fragment) for target cisco in policy
 keywords (fragment_offset) for target cisco in policy
 keywords (is-fragment) for target iptables in policy
 keywords (loss_priority) for target arista in policy
 keywords (loss_priority) for target ciscoxr in policy
 keywords (owner) for target pcap in policy
 keywords (policer) for target iptables in policy
 keywords (protocol_except) for target iptables in policy
 keywords (protocol_except) for target speedway in policy
 keywords (qos) for target iptables in policy
 keywords (qos is-fragment) for target iptables in policy
 keywords (qos policer) for target iptables in policy
 keywords (tcp-initial) for target arista in policy

This is a short list I found of things that would break for us. A lot of these are used on other platforms also. I would hate to accept this and miss out on some very obvious ones like QOS and break others using Capirca.

sneakywombat:
@ankenyr, now that is more like it. :)  I was really surprised when you came back and said that there were no conflicts with your policy terms.  I had fully expected this to break some terms.  

What I'd like to know is if the tokens I said were unsupported, were mistakenly omitted in my code mods.  It is totally possible I overlooked some.  We'd need to look through these platforms to know for sure though.  For example, the (owner) pcap one, I believe it is accurate in that the owner token really isn't supported in the pcap lib.  If you look in the source, there is no mention of it.  Whereas, if you look at another platform like cisco, https://github.com/google/capirca/blob/master/lib/cisco.py#L422-L423, you'll see it clearly used.  I see no such use like this in pcap.py.

When I created the supported tokens and supported sub-tokens data structures, I used a number of different grep searches on the platform files to find what tokens were used.  I also did a very tedious stare and compare effort to follow it up.  Again, it's totally possible that I missed some, but at least this one example, is a perfect example of what this pr is trying to prevent - people writing terms for a platform target and expecting (or assuming) the platform target to implement all of the tokens that the term contains.

ankenyr:
Owner and expiration should be supported everywhere. Even if there is nothing rendered in the output it is useful to have it. We scan our pol files and notify owners prior to the expiration of terms. I am sure others do similar things. So I would make sure that those are in every supported field for every platform.

I need to take a look at the ones that are failing a bit closer. It could be that we are not using platform-exclude properly and just relying on the fact that something doesn't do anything. Will get back to you tomorrow but please make sure owner and expiration is included in every platform.

sneakywombat:
That's it totally reasonable in the owner case. We could have a couple be
no op tokens, like in the case of owner. Expiration should totally be
supported and implemented though.

ankenyr:
I agree but I know it is not in at least Aruba and I would suspect any generator not made by a member of my team may not include it. I or someone else would need to check just to make sure. This should be done as a follow up and should not be seen as a blocker to your submitting

This is really great though as we can easily see what things are supported in a much easier fashion. A small tool could be made to collect these and make maintaining documentation much easier.

Looked at a couple examples and it seems that the problem has to do with platform and possibly platform-exclude.

Take the following example

```
header {
  comment:: "foobar"
  target:: cisco foobar
  target:: juniper foobar
}

term accept-first {
  destination-address:: INTERNAL
  protocol:: tcp
  action:: accept
}

term accept-second {
  destination-address:: INTERNAL
  fragment-offset:: 1-2
  platform:: juniper
  action:: deny
}

term accept-third {
  destination-address:: INTERNAL
  fragment-offset:: 1-2
  action:: accept
}
```

I get the following output

```
<class 'capirca.lib.aclgenerator.UnsupportedFilterError'>
 accept-second contains unsupported keywords (fragment_offset) for target cisco in policy 
accept-third contains unsupported keywords (fragment_offset) for target cisco in policy 
```
I would expect this to be true for accept-third but not for accept-second. This should have a check to make sure it skips testing excluded terms.

sneakywombat:
interesting case with the fragment-offset.  Since you're calling platform:: on the term, it should skip it.  I did accidentally remove some code in aclgenerator, which I put back last night in an update.  Can you look to see if the aclgenerator init code you have from your testing includes this:

```
          if term.platform:    
            if self._PLATFORM not in term.platform:   
              continue    
          if term.platform_exclude:   
            if self._PLATFORM in term.platform_exclude:   
              continue
```

ankenyr:
Well that is embarrassing, looks like this went away. I guess I last synced before that. Will rerun my tests

ankenyr:
Ok so there are still a lot of things left over. Things like qos, counters, logging, and some others (still trying to get an exhaustive list). There seems to be a common theme to these which is what I feel a lot of the warnings and errors common throughout the generators were supposed to do.

Certain keywords on every platform are supported which is something your change is great at pointing out. There are also a certain set of keywords that while they are not supported are safely ignored. It has been the intent of Capirca to enforce policy across platforms as best as it can.

In the case of the following policy 

header {
 target:: cisco foo
 target:: juniper foo
}
term bar {
  destination-address:: MAIL_SERVERS
  destination-port:: MAILSERVICES
  protocol:: tcp
  action:: accept
  qos: baz
}

What happens and what should happen?
In this case it would cause an error but should it? I think in cases where it will not cause a conflict in policy enforcement it is fine to make a warning and continue. Otherwise you need to start messing with platform and platform-exclude and possibly having multiple almost duplicate terms.

What is your and others opinions on this?

sneakywombat:
Hey @ankenyr, 

I understand the intent of capirca to enforce policy as best it can, across as many platforms as possible.  I think that's awesome, but to me, it comes at a cost.  I'm in the camp that platforms should fail if someone passes a token that it doesn't implement.  That error is easy to communicate to development from an end user.  We've found that end users are more often surprised that a given platform wasn't doing what they thought it was doing based on the tokens they used, because platforms can sometimes eat the tokens and do nothing with them.  It's also very difficult to keep up with what a platform implements and what it will silently ignore.  Also, each platform warns about "accepted, but not implemented" tokens with varying levels of thoroughness.  Specifically, your comment echoes this: 

> I feel a lot of the warnings and errors common throughout the generators were supposed to do.

Nevertheless, I concede that there should be some wiggle room with what happens here.  I started with the maximum level - fail hard.  I wanted to see if you guys would bite on that. :)  However, as you have explained in your most recent comment about the intent of capirca and the desire to render as many platforms as possible; I totally understand.  I actually wondered about that myself when I was writing all of this, so it's good to have that voiced so I know for sure.

So, how about this instead... if we rely on a platform to implement a set of warnings about tokens it may or may not implement, we'll be chasing our tail on changes and new platforms.  I think these warnings should be centralized in the parent class and applied equally to all subclasses.  Doing so takes the burden of creating these warnings off of the platform implementation and makes the warnings identical for all platforms since they are implemented in one place, consistently.  For the cases that you have cited above, expiration, etc, do you have a firm list of tokens that all platforms *should* implement?  I remember reading this in the wiki, but I'd like to firm those up as correct.  We can warn on those in the parent class, just like we do for the supported tokens/supported subtokens.  EG, Any tokens passed that are not explicitly marked as supported by the supported tokens/subtokens code would warn one time, should they intersect with this "should be implemented" set, but not cause a failure in rendering.

If I make a simple example to illustrate what i'm thinking here, and what it sounds like you want -

parent class:
should_be_supported_tokens = {expiration, owner, logging}
supported_tokens = {expiration, owner, logging}
supported_subtokens = {...}

platform subclass:
supported_tokens = {expiration, owner}
supported_subtokens = {...}

Then at runtime, aclgenerator._init will collect the platforms implemented tokens and determine if any should_be_supported_tokens were not implemented.  If so, it can warn then, but proceed anyway.  In the above example, we'd warn on logging not being implemented, but still proceed with rendering. This is different from the functionality in this diff now in that it will fail with an exception.  Would that be a good compromise here?  

In the end, as you have also identified, I would love to know for sure what a platform implements and not have to dig in the code to find it when someone asks. It would benefit documentation and also error conditions that may not be known (a platform assumed to do something that it doesn't).  I agree with you though that i may have been heavy handed with the hard failure :)  It's just how I tend to code things.  If you are ok with this should_be_supported_tokens approach, I can update this pr with those changes.

Thanks for working through this with me!

ankenyr:
The name "should be supported" is a little bit of a misnomer. These are keywords which do not affect the policy enforcement but which may or may not be rendered. It is not a matter that they should be supported since qos is not really applicable to the Cisco ACLs but the ACL should not error out. Either case the names can always be updated if we can think of something better. Maybe we can just rely on documentation to explain this.

I think we are on the same page though, there are basically a list of keywords that can be used which do not affect policy. Warning instead of erroring out in these cases gives us the flexibility to move forward. I would expect that others starting out would modify these lists  to be more strict if they want and we should probably provide a bit of documentation on the wiki to call this out. I will look to you for help on that once we make pull in the changes.

If you feel we are on the same page let me know. This looks like the list of changes that need to be made to make sure we are able to still compile. Some things will need to be added which should not remain. We will want to mark these with comments saying something akin to "REMOVE ME" and we will want to make a issue to track their removal.

common keywords to ignore or to be added to should_be_supported (what ever name is decided on)
                        'policer',
                        'qos',
                        'counter',
                        'logging',
                        'owner',
                        'loss_priority',

iptables:
add is-fragment to options, need to check all places this is used in conflict to resolve

pcap.py:
add established and is-fragment to options, need to either add functionality if it exists or resolve this conflict

packetfilter:
add is-fragment to options, need to check all places this is used in conflict to resolve


cisco:
add reject to actions, currently it just maps that to deny, we have a few places that use this and need to be fixed.

sneakywombat:
sounds good to me.  I'll make the adjustments.  As for the name, would
"warn_if_unsupported" be better?

-j

ankenyr:
That sounds like a good name. Thanks @sneakywombat !

sneakywombat:
putting in an update now, but as for the cisco behavior of:

> cisco:
> add reject to actions, currently it just maps that to deny, we have a few places that use this and need to be fixed.

We have always thought that to be harmful behavior - providing an action and having it get silently re-written. I'll add this back, but would like to discuss it outside of this pr.

ankenyr:
I don't think we need a discussion. I am on the same page and think it is harmful. I would need to fix the instances internal and once those are fixed we can remove that entry.

sneakywombat:
@ankenyr, I looked at the warnings module a bit more, to err on the side of caution, i can convert these to instead log via the logging module.  Will update tomorrow.

ankenyr:
Thanks for looking into that. Did you find possible complications that could arise from that modules use?

sneakywombat:
@ankenyr i fixed the action support for the cisco line.  Let me know how testing goes.

ankenyr:
Hey @sneakywombat,
I think we are getting close to accepting this! Just a couple small things.

Looks like warnings is still in use on pcap_test, was this intended?

I see on ipset you made its own list. It really should share the same list as iptables right? The first step in the str() function has it calling iptables string.

Add 'routing_instance' to the WARN_IF_UNSUPPORTED list. I believe this should be treated similar to the 'counters' and 'logging'

Looks like it is assumed that there is a filename in policy. This isn't necessarily true if you were to generate the policy in some other way and feed it into a generator. Lets at the minimum add this attribute as a blank string at policy's initialization. So somewhere around https://github.com/google/capirca/blob/master/lib/policy.py#L158 please add a line saying

`self.filename = ''`

sneakywombat:
I'll take a look this weekend and update. Good finds. The warnings mod
definitely was missed in my revert.

sneakywombat:
woohoo! @ankenyr I believe i'm on the hook for documentation.  anything specifically you wish?

ankenyr:
Lets continue the discussion in a issue instead of this pull request since it is closed.
I would like all current documentation in the wiki. The files in the doc directory should go away.

For documenting everything we should work on https://github.com/google/capirca/wiki/Policy-format
I would like to keep a similar format. There should be a section describing the behavior of aclgenerator and the keywords supported by every platform (source-address, source-port, action etc) and the options that every platform should include (tcp established etc).

Each platform will have its own section which will document its header information like it has now plus anything that differentiates it from the general behavior (juniper has is-fragment, aruba having limited support of tokens)

I am not sure how you want to collaborate on this though as the wiki's don't support pull requests. I looked up a few "work arounds" but none of them looked simple. I am happy to entertain any ideas.

PR #97, opened 2016-09-23
Add support for built-in POSTROUTING chain to iptables generator

rvineyard:


rvineyard:
I should probably add PREROUTING while I'm at it...

rvineyard:
Actually this is going to need a bit more work before the output is iptables-restore compliant again. PREROUTING and POSTROUTING need to go under the *nat table instead of the *filter table

PR #98, opened 2016-09-23
fix broken characterization tests

sneakywombat:
fix MOE_MIGRATED_REVID=131342769, it broke all characterization tests that utilized the cisco.py platfrom generator.


ankenyr:
Thanks for fixing these ones so quickly Joe!

PR #99, opened 2016-10-04
Add IndentList and fix some pylint style messages

bc-23:
Add a new class IndentList instead of normal list which takes care of the
required indent syntax.

This is only a change to the style, the result output stays the same.

Signed-off-by: Florian Zavatzki fz@binsec.com

### DIFF
--- a/lib/junipersrx.py
+++ b/lib/junipersrx.py
@@ -28,19 +28,6 @@ from lib import nacaddr
 import logging
 
 
-def junipersrx_list(name, data):
-  return '%s [ %s ];' % (name, ' '.join(data))
-
-
-class IndentList(list):
-  def __init__(self, indent, *args, **kwargs):
-    self._indent = indent
-    super(IndentList, self).__init__(*args, **kwargs)
-
-  def iappend(self, size, data):
-    self.append('%s%s' % (self._indent * size, data))
-
-
 class Error(Exception):
   """generic error class."""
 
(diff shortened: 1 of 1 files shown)


bc-23:
I signed it!

bc-23:
I signed it!

ankenyr:
Hi @bc-23,

```
Author: removed <removed>
Date:   Tue Oct 4 14:39:27 2016 +0200

    Add IndentList and fix some pylint style messages
    
    Add a new class IndentList instead of normal list which takes care of the
    required indent syntax.
    
    Signed-off-by: removed <removed>

```

I marked the name and email as removed but if you do a `git log` yourself you will see that the email doesn't match what you signed with.

You have a couple options at this point
1) Possibly amend the commit to use the email you signed the CLA with
https://www.atlassian.com/git/tutorials/rewriting-history/git-commit--amend
2) Sign the CLA with the email already used
3) Delete PR, update email, redo commit, make a new push request
https://help.github.com/articles/setting-your-email-in-git/

bc-23:
Hi,

yes, I signed a CLA yesterday for the company, but signing a CLA for a company seems to be more complicated. As far as I can tell it is now pending on google for signing.

When I don't get the CLA today I will update the commit mail address.

lib/junipersrx.py
```diff
@@ -28,6 +28,19 @@
 import logging
 
 
+def junipersrx_list(name, data):
+  return '%s [ %s ];' % (name, ' '.join(data))
+
+
+class IndentList(list):
+  def __init__(self, indent, *args, **kwargs):
```
ankenyr: Can you insert a line before this def?

ankenyr: This should also go down under the errors.

bc-23:
I moved the IndentList class below the error definitions and added a newline before the def

lib/junipersrx.py
bc-23: I moved the IndentList class below the error definitions and added a newline before the def

PR #102, opened 2016-10-26
Disable state module for most ICMPv6 types

bc-23:
The conntrack module for ICMPv6 only supports NEW states for type 128 and 139.
Using the state module for e.g. 135 will prevent the rule from matching and
results in a not working ND.

Signed-off-by: Florian Zavatzki fz@binsec.com

### DIFF
--- a/lib/iptables.py
+++ b/lib/iptables.py
@@ -444,7 +444,11 @@ class Term(aclgenerator.Term):
       options.append('-m u32 --u32 "0x3&0xff=0x2c"')
 
     # set conntrack state to NEW, unless policy requested "nostate"
-    if self.trackstate:
+    # for icmpv6 use trackstate only for type 128 and 139, as this is the
+    # only supported types in nf_conntrack_proto_icmpv6
+    if self.trackstate and (
+        protocol != 'icmpv6' or icmp_type in [128, 139]
+    ):
       already_stateful = False
       # we will add new stateful arguments only if none already exist, such
       # as from "option:: established"


lib/iptables.py
```diff
@@ -444,7 +444,11 @@ def _FormatPart(self, protocol, saddr, sport, daddr, dport, options,
       options.append('-m u32 --u32 "0x3&0xff=0x2c"')
 
     # set conntrack state to NEW, unless policy requested "nostate"
-    if self.trackstate:
+    # for icmpv6 use trackstate only for type 128 and 139, as this is the
+    # only supported types in nf_conntrack_proto_icmpv6
+    if self.trackstate and (
+        protocol != 'icmpv6' or icmp_type in [128, 139]
```
rvineyard: might want to add another line to cover the same situation in ipv4:

    protocol != 'icmp' or icmp_type in [ 8, 13, 15, 17 ]

vklimovs:
ICMP state tracking has no issue dealing with state for any type:

pulley ~ # iptables -A INPUT -i br0 -p icmp -m icmp --icmp-type 8 -m conntrack --ctstate NEW -j ACCEPT
pulley ~ # iptables -A INPUT -i br0 -p icmp -m icmp --icmp-type 7 -m conntrack --ctstate NEW -j ACCEPT
pulley ~ # iptables -A INPUT -i br0 -p icmp -m icmp --icmp-type 31 -m conntrack --ctstate NEW -j ACCEPT

all succeed.

rvineyard:
vklimovs all iptables commands succeed, yes, however if you look at the kernel netfilter code you'll see that only certain message types actually result in state tracking for a new connection. See http://lxr.free-electrons.com/source/net/ipv6/netfilter/nf_conntrack_proto_icmpv6.c#L121 for the relevant ipv6 implementation and http://lxr.free-electrons.com/source/net/ipv4/netfilter/nf_conntrack_proto_icmp.c#L110 for the same thing in ipv4.

vklimovs:
Ah, I see. Good catch, Robert!

PR #109, opened 2016-12-18
add __future__ to all python files to support python2.7/3 code

sneakywombat:
Summary:

I'd like to add these future imports to all of the python code.
This will allow us to write python3 compatible code more easily, with the
eventualy hope that we can support python3 in the future.  The last big
hold out for not doing this was python-gflags not supporting python3. That
changed sometime last month so i'd like to get us moving in that direction.
The end is near.  https://pythonclock.org/

The harder part of this effort revolves around nacaddr.  We've solved this
once internally, but it could use some more work.  That effort will be in
later PR requests.

Test Plan: python2.7 -m unittest discover -p *_test.py

......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Ran 518 tests in 82.943s

OK

### DIFF
--- a/aclcheck_cmdline.py
+++ b/aclcheck_cmdline.py
@@ -15,6 +15,11 @@
 
 """Command line interface to aclcheck library."""
 
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
 __author__ = 'watson@google.com (Tony Watson)'
 
 from optparse import OptionParser
--- a/aclgen.py
+++ b/aclgen.py
@@ -18,6 +18,11 @@
 
 """Renders policy source files into actual Access Control Lists."""
 
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
 __author__ = 'pmoody@google.com'
 
 import copy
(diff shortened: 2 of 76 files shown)


sneakywombat:
I forgot to add, we don't need this PR at all if we just move to py3.  I took the dual-support path though as I assumed that would be the most desirable.

ankenyr:
Taking a look at this now, what was wrong with the CLA?

sneakywombat:
I had some .gitconfig settings wrong. Issue was in my end, all better. :)

ankenyr:
Mind moving the imports under the author tags in the files?

sneakywombat:
I can. Is that preferred? I tried to follow this,
https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names. The
author tags tripped me up, but ultimately gave precedence to the future
imports.

ankenyr:
huh, guess that is how it should be. Sorry! This looks fine to me how it is then. Need to check a few more things but should be able to accept this.

ankenyr:
Hey @sneakywombat, this looks good but there are a few tests that will break internal if I pull this in. Running out of time today but will pick this up tomorrow when I can fix those tests at the same time of the merge. Thanks for doing this.

sneakywombat:
Great! Thanks!

ankenyr:
Thanks for the work @sneakywombat

Issue #62, opened 2016-05-31
SRX security zone VPN not accepted


XioNoX:
I'm testing your latest commits with my policies.

The policies containing for example:
"target:: srx from-zone corp to-zone vpn"
fail because of the vpn zone:
```
Traceback (most recent call last):
  File "aclgen.py", line 498, in <module>
    main(FLAGS(sys.argv))
  File "aclgen.py", line 461, in main
    FLAGS.exp_info, write_files)
  File "aclgen.py", line 193, in RenderFile
    input_file, sys.exc_info()[0], sys.exc_info()[1]))
__main__.ACLParserError: Error parsing policy file policies/pol/offices_srx_base_policies-global.pol:
<class 'lib.policy.ParseError'> ERROR on "vpn" (type VPN, line 262, Next '}')
```
If I rename vpn into something else it then works fine.

ankenyr:
I believe I see the problem. vpn is a option inside of terms.
Try changing it to vpn_s and it works, then try changing it to something else that is a term option like comment and it breaks.

I will need to explore the parsers a bit to see if there is a way to search for begining of lines.

In the mean time you can easily get around this by renaming it from vpn to something like vpn_zone or remote_access or some other name besides "vpn"

PR #113, opened 2017-01-18
Add Palo Alto Firewalls support

XioNoX:
This PR includes the content of the PR https://github.com/google/capirca/pull/52
Plus the following modifications:
* Generic address book and address sets
* PAN application (> layer 4) with pan-application:: xxx yyy
* Raises errors if policies/services names are too long
* uses service_name and protocol in service name instead of port list
* Used autopep8 (atom beautify) on paloaltofw.py

If this gets accepted, we also need to make sure the original author of PR #52 gets the credits as he did the largest part of the work.


aclgen.py
```diff
@@ -54,6 +54,7 @@
 from lib import speedway
 from lib import srxlo
 from lib import windows_advfirewall
+from lib import paloaltofw
```
ankenyr: Please move this to be in alphabetical order.

lib/aclgenerator.py
```diff
@@ -225,6 +225,7 @@ class ACLGenerator(object):
                          'owner',
                          'qos',
                          'routing_instance',
+                         'pan-application'
```
ankenyr: Forgot a comma?
Also is this strictly necessary? I don't believe it is.

lib/policy.py
```diff
@@ -1600,6 +1617,7 @@ def __ne__(self, other):
     'protocol': 'PROTOCOL',
     'protocol-except': 'PROTOCOL_EXCEPT',
     'qos': 'QOS',
+    'pan-application':'PAN_APPLICATION',
```
ankenyr: Can we put a space after the :?

lib/policy.py
```diff
@@ -1612,7 +1630,7 @@ def __ne__(self, other):
     'timeout': 'TIMEOUT',
     'traffic-type': 'TRAFFIC_TYPE',
     'verbatim': 'VERBATIM',
-    'vpn': 'VPN',
+#    'vpn': 'VPN',
```
ankenyr: Was this meant to be commented out? I don't see why this should be.

lib/policy.py
```diff
@@ -2080,6 +2099,11 @@ def p_qos_spec(p):
   """ qos_spec : QOS ':' ':' STRING """
   p[0] = VarType(VarType.QOS, p[4])
 
+def p_pan_application_spec(p):
```
ankenyr: two spaces should be separating this def, same thing for under it.

lib/policy_simple.py
```diff
@@ -249,6 +249,8 @@ class ProtocolExcept(Field):
 class Qos(Field):
   """A rate-limit-icmp field."""
 
+class PANApplication(Field):
```
ankenyr: Same here, please use two spaces between these classes.

lib/paloaltofw.py
```diff
@@ -0,0 +1,585 @@
...
+"""Palo Alto Firewall generator."""
+
+import collections
+import datetime
+import logging
+
+import aclgenerator
+import nacaddr
+
+import re
```
ankenyr: This looks to be unused.

XioNoX:
I signed it! Corporation Mozilla.

aclgen.py
XioNoX: Done

lib/aclgenerator.py
XioNoX: Indeed, I wasn't sure of its purpose when I originally added it.

lib/policy.py
XioNoX: Done

lib/policy.py
XioNoX: Indeed no. This is a workaround specific to our setup, as vpn is a zone name and not a special keyword.
Cf. https://github.com/google/capirca/issues/62
I removed the comment sign.

lib/policy.py
XioNoX: Done

lib/policy_simple.py
XioNoX: Done

lib/paloaltofw.py
XioNoX: Indeed, removed.

lib/paloaltofw.py
```diff
@@ -0,0 +1,583 @@
...
+            "-".join(self.options["to_zone"]) + "-" + term.name
+# TODO: uncomment when ready to use
+#    if len(rule_name.decode("utf-8")) > 31:
+#        raise PaloAltoFWTooLongName('Rule name must be 31 characters max: %s'
+#                                        % str(rule_name))
+        self.rules[rule_name] = self.options
+
+
+class PaloAltoFW(aclgenerator.ACLGenerator):
+    """PaloAltoFW rendering class.
```
ankenyr: This docstring should only be one line. Arguments should not be documented here and the description below is kind of redundant. It inherits from the aclgenerator class so it should be obvious it renders a syntax output.

policies/pol/sample_paloalto.pol
```diff
@@ -0,0 +1,48 @@
+################
```
ankenyr: Can we see an example of pan_application being used inside of here?

lib/paloaltofw.py
```diff
@@ -0,0 +1,583 @@
...
+
+            if filter_type not in self._SUPPORTED_AF:
+                raise UnsupportedHeader(
+                    'Palo Alto Firewall Generator currently does not support %s as a header option' %
+                    (filter_type))
+
+            term_dup_check = set()
+            new_terms = []
+            for term in terms:
+                term.name = self.FixTermLength(term.name)
```
ankenyr: Not a big deal but is there a max term length? If so we should probably override it because it is likely different from 62 limit in aclgenerator.

lib/paloaltofw.py
```diff
@@ -0,0 +1,583 @@
...
+    def __init__(self, from_zone, to_zone, terms):
+        # Palo Alto Firewall rule keys
+        self.options = {}
+        self.options["from_zone"] = [from_zone]
+        self.options["to_zone"] = [to_zone]
+
+        self.modify_options(terms)
+
+    def modify_options(self, PAFWterm):
+        # for PAFWterm in terms:
```
ankenyr: Should this be commented out?

lib/paloaltofw.py
```diff
@@ -0,0 +1,583 @@
...
+
+        final_service_name = 'service-' + service_name + '-' + protocol
+        if len(final_service_name.decode('utf-8')) > 63:
+            raise PaloAltoFWTooLongName(
+            'Service name must be 63 characters max: %s'
+                                        % str(final_service_name))
+        self.service_map[(ports, protocol)] = {'name': final_service_name}
+
+
+class Rule():
```
ankenyr: I am confused why this exists, it seems to replace the function of Term. Why was Term not able to be used instead?

lib/paloaltofw.py
XioNoX: Updated, I think it has been copied from the JuniperSRX docstring, which is very similar.

policies/pol/sample_paloalto.pol
XioNoX: Added, as well as the expected output in the test folder.
Palo alto comes with a wide array of existing "Applications", that's why I made it a simple variable that the user can define.

lib/paloaltofw.py
XioNoX: I'm not sure I understand.
The final rule length is limited to 31 chars. I added an exception for that on line 208  (PaloAltoFWTooLongName)

lib/paloaltofw.py
XioNoX: I believe the loop has been moved to line 346 (for term in new_terms:)
I removed the line mentioned here.

lib/paloaltofw.py
XioNoX: Not sure why the original Author went that way.

policies/pol/sample_paloalto.pol
ankenyr: Can someone make custom applications or is there a set list of things? If it is a set list should this have checking to make sure someone is not applying incorrect services?

lib/paloaltofw.py
ankenyr: There is a variable named [_TERM_MAX_LENGTH](https://github.com/google/capirca/blob/master/lib/aclgenerator.py#L265). If there is a max length FixTermLength should be called and _TERM_MAX_LENGTH should be set. This will raise an error if the term name is to long. If there is no limit then FixTermLength is irrelevant. Make sense?

lib/paloaltofw.py
```diff
@@ -0,0 +1,579 @@
...
+        rules.append(self.INDENT * 10 + "<member>any</member>")
+      else:
+        for s in options["service"]:
+          rules.append(self.INDENT * 10 + "<member>" + s + "</member>")
+      rules.append(self.INDENT * 9 + "</service>")
+
+      rules.append(self.INDENT * 9 + "<action>" + Term.ACTIONS.get(
+          str(options["action"])) + "</action>")
+
+      if fz == tz == "any":
```
ankenyr: This seems weird. There are loops above over the from zone and to zone which is a list. From my bit of testing I don't see that this would ever have more than one from-zone or to-zone per rule in rules. So why are these added as a list? The way the logic is setup fz could theoretically be set to an empty list if someone fed in a modified policy object that had the options set to ['from-zone', '', 'to-zone', ''] this would cause a problem.

policies/pol/sample_paloalto.pol
XioNoX: It's also possible to make custom applications.

lib/paloaltofw.py
XioNoX: Currently the final name (limited at 31 chars) also contains the source and destination zone name (eg. internal_2_external-term-name. So only checking the term length is not possible (and ends up being quite short).
In order to save rule length, I'm considering the idea of requiring the terms to be unique (similar to the srx lib for example), and then either:
1/ Trimming the sources zone names to 3 char (eg. int_2_ext-term-name), then the prefix is always a fixed length.
or
2/ Not applying any prefix, and letting the user manually set a prefix (eg, src/dst zones) in the term if the user wants it.
I think I like 2/ better as it lets the user have more freedom.
What do you think?

lib/paloaltofw.py
XioNoX: @ankenyr
Not sure why it's originally set up that way, but it can be related to the fact that PAN allows rules to have several source/destination zones.
Right now line 149 
```
self.options["from_zone"] = [from_zone]
```
sets it to an array containing only 1 element.
But we could for example imagine it (in the future) being a list from the .pol file header, line 292
```
self.from_zone = filter_options[1].split(',')
```
(or something similar).

About the possibility of fz or tz being empty, I believe we can add a check after line 150:
```
    if not from_zone or not to_zone:
        raise PaloAltoFWOptionError("Source or destination zone is empty.")
```

XioNoX:
CLA signed with my personal email as well.

lib/paloaltofw.py
ankenyr: 2 gives more freedom but there should be checks to make sure a invalid ACL is not rendered. Please make sure that is not possible via raising an error.

lib/paloaltofw.py
ankenyr: Can we explore why this was chosen and if this could be rolled up into a term? I don't understand why this is being used this way and I don't like having code integrated that no one understands. I will expect a docstring for Rule explaining what it is and why it is used instead of the term object.

lib/paloaltofw.py
XioNoX: Added a check to raise "PaloAltoFWDuplicateTermError" in case a term is duplicated.

XioNoX: As well as use _TERM_MAX_LENGTH instead of custom check.

lib/paloaltofw.py
XioNoX: I'm going to need help to figure that one out.

XioNoX:
Commit https://github.com/google/capirca/pull/113/commits/48a5f7120243610ca62fb92d3951e126fabe10c8 default the "service" field to "application-default" if "service" is not set, but "application" is.
If neither "service" nor "application" is set, the default "any" is used.

This is to follow PaloAlto recommendations and improve security. If "any" is used, the "application" filter can in some cases not be respected.

Here is PaloAlto suport take on it:
> When the first packet is coming into firewall, the firewall rules are evaluated with 6 tuples (Source/Destination IP address, Source/Destination Ports, Protocol, and Zone). The application is not necessarily known in this first packets, it can take several packets to determine what the underlying application is. During this evaluation period, packets may be allowed through if there is a rule which has "any" instead of "application-default" for service or an application using dynamic tcp/udp ports. In this case you have a rule with "any" service it means that all ports will be allowed till application is identified.
> After the session established with 6 tuples the application will be identified and it will be ultimately blocked by rule set accordingly.
> 
> If you change it to "application-default", firewall will only allow standard ports for applications you configured in a rule when the session is set up. So in security aspect we recommend to use "application-default" for service.
> 
> Please refer to KB below for more detail.
> https://live.paloaltonetworks.com/t5/Management-Articles/Why-are-Rules-Denying-Applications-Allowing-Some-Packets/ta-p/58098
> https://live.paloaltonetworks.com/t5/Management-Articles/URL-Block-Page-Response-Page-Appears-When-Custom-URL-is-Used/ta-p/56107

ankenyr:
This is looking good enough that I think you should start making tests for the PR.

XioNoX:
Great! Tests added!

lib/policy.py
```diff
@@ -1026,6 +1040,7 @@ def AddObject(self, obj):
       # qos?
       elif obj.var_type is VarType.QOS:
         self.qos = obj.value
+
```
ankenyr: Extra space added, please remove.

lib/policy.py
```diff
@@ -1334,6 +1349,7 @@ class VarType(object):
   HOP_LIMIT = 47
   LOG_NAME = 48
   FLEXIBLE_MATCH_RANGE = 49
+  PAN_APPLICATION = 50
```
ankenyr: Can we change this to 54? I have some internal stuff I will be pushing out soon and it will conflict.

lib/policy.py
XioNoX: Removed!

lib/policy.py
XioNoX: Changed!

PR #127, opened 2017-03-29
Fix source-exclude/destination-exclude keywords in SRX generator

pw378:
Fix issue where source-exclude and destination-exclude keywords 
generated unexpected output of null address book names.  In addition, 
the use of a naked *-exclude raised errors.

This change leverages the term name to create an address-book name
for the IPs resulting from excluding source or destinations.

Appropriate unit tests are also included.

Added a README in ./tests/ for reference.


ankenyr:
Thanks for the work on this @pw378!

PR #134, opened 2017-06-09
Changing NSX Capirca Generator Output to xml format.

rarcotvmw:


ankenyr:
Looks like tests are not passing.
Also why did you decide to use lxml instead of xml

rarcotvmw:
xml.dom.minidom adds an extra white space to the xml.
xml will be like this...
<section id="1234" name="sample_nsxv_filter">
	<rule logged="false">
		<name>accept-icmp</name> ....
It adds 2 spaces instead of 
<section id="1234" name="sample_nsxv_filter">
  <rule logged="false">
    <name>accept-icmp</name> ...

rarcotvmw:
<section id="1234" name="sample_nsxv_filter">
	<rule logged="false">
instead of 
<section id="1234" name="sample_nsxv_filter">
  <rule logged="false">

rarcotvmw:
If the extra white space is okay, I can use xml.dom.minidom.
Please let me know.
Thank you,
Rajitha

ankenyr:
Will it affect the ACL? Will it be able to be committed? If you pull the ACL back out of the system does it still contain that space or would it cause a diffing problem?

rarcotvmw:
No it does not affect the ACL. Yes it can be committed. If we pull the ACL back out of the system, the space would still be there. The REST call with NSX manager will go through fine. 
Did not understand about the diffing problem.

Also, I had created another pull request https://github.com/google/capirca/pull/135 which uses  xml.dom.minidom

ankenyr:
I am closing this #135 is the current pull and there doesn't seem to be anything here that isn't taken care of by that PR.

PR #135, opened 2017-06-09
Changing NSX Capirca Generator Output to xml format using xml.dom

rarcotvmw:
Creating this request using xml.dom.minidom


lib/nsxv.py
```diff
@@ -614,10 +615,11 @@ def __str__(self):
...
-      target.append('\n')
 
       # ensure that the header is always first
       target = target_header + target
       target.append('%s' % (_XML_TABLE.get('sectionEnd')))
       target.append('\n')
-    return '\n'.join(target)
+
+      target_as_xml = xml.dom.minidom.parseString(''.join(target))
+    return target_as_xml.toprettyxml()
```
SecurityForDotes: I would add the option indent="  " to the toprettyxml(), that removes the tab and makes it two spaces. (taken from: https://docs.python.org/2/library/xml.dom.minidom.html)

tests/characterization_data/filters_expected/sample_nsxv.nsx
```diff
@@ -1,54 +1,612 @@
-<!--
- $Id:$
- $Date:$
- $Revision:$
--->
+<?xml version="1.0" ?>
```
SecurityForDotes: What caused this new header to get implemented?

tests/characterization_data/filters_expected/sample_nsxv.nsx
```diff
@@ -1,54 +1,612 @@
...
+<!-- $Id:$ $Date:$ $Revision:$-->
```
SecurityForDotes: The new lines were lost for these keywords.  We need to keep that newline structure.

lib/nsxv.py
rarcotvmw: Will make the change.

tests/characterization_data/filters_expected/sample_nsxv.nsx
rarcotvmw: Will make the change.

tests/characterization_data/filters_expected/sample_nsxv.nsx
rarcotvmw: xml.dom.minidom toprettyxml adds the xml header by default.

lib/nsxv.py
rarcotvmw: Done

tests/characterization_data/filters_expected/sample_nsxv.nsx
rarcotvmw: Done.

tests/characterization_data/filters_expected/sample_nsxv.nsx
ankenyr: You can remove this I believe. You would need to serialize the xml to a string using the [toxml function](https://docs.python.org/2/library/xml.dom.minidom.html#xml.dom.minidom.Node.toxml)

@SecurityForDotes is this header a problem or is it just a nit?

rarcotvmw: Actually, the document created by xml.dom.minidom.parseString() has the xml header by default. I think xml.dom.minidom.toprettyxml does not have the option of removing the xml header.
Does the xml header cause a problem? Please let me know.

SecurityForDotes: @ankenyr a nit to the best of my knowledge (in that I didn't try to load it to see what would happen).

I'll load it into NSX and see if it gets returned in the same way.

SecurityForDotes: Looks like it is fine to have, if it is not there it gets added when you extract it.

lib/nsxv.py
```diff
@@ -614,10 +616,11 @@ def __str__(self):
...
-      target.append('\n')
 
       # ensure that the header is always first
       target = target_header + target
       target.append('%s' % (_XML_TABLE.get('sectionEnd')))
       target.append('\n')
-    return '\n'.join(target)
+
+      target_as_xml = xml.dom.minidom.parseString(''.join(target))
+    return target_as_xml.toprettyxml(indent=" ")
```
SecurityForDotes: I meant to put in 2 spaces here. I think it looks better than one for readability.

However as rob has stated we definitely prefer to match the actual output that NSX provides when you pull the ACL. I have not been able to find out what format I actually get back from NSX, I've gotten different formats based on the tooling I use to query. Do you know what NSX will actually provide back?

rarcotvmw: Added 2 spaces indentation.
NSX provides the header tag  "  xml version="1.0" encoding="UTF-8" ".

rarcotvmw: Would you prefer the xml header as " xml version="1.0" encoding="UTF-8" "? Please let me know.

rarcotvmw: Or would you like to suppress the header? I can make the change accordingly.

SecurityForDotes: I would prefer that the header matches that NSX does when it gets extracted. So if NSX adds the encoding lets get it added. If this currently matches the NSX output we can leave it.

rarcotvmw: Added encoding="UTF-8" to xml header. The header now is the same as what NSX provides.

Issue #137, opened 2017-06-29
NSXV generator inet_version mixed outputs invalid mix of IPv6 <-> IPv4 (when only thing on each side)


SecurityForDotes:
The nsxv generator behaves incorrectly when inet_version mixed (in the header) is used and the source and dest have opposite inet_version for the terms.

Example:
if SourceAddress contains only IPv4 addresses
if DestinationAddress contains only IPv6 addresses

Then currently it outputs an ACL, when in fact it shouldn't because that's not a valid combination (same if source and dest are flipped.

I  think some valid combinations are (non exhaustive):
“any” to IPv6 
IPv6 to “any”
IPv6 to IPv6
IPv6+IPv4 to IPv6+IPv4
“any” to IPv4
IPv4 to “any”
IPv4 to IPv4
IPv4 to IPv4 +IPv6 (results in IPv6 being dropped from term and vice versa)
IPv6 to IPv6 +IPv4 (results in IPv4 being dropped from term and vice versa)

IE: it has to be balanced, and matched.

Given the following cases It looks like a number of them are obviously wrong:
mixed-to-v6
mixed-to-v4
mixed-to-mixed
v4-to-mixed
v4-to-v4
v4-to-v6
v6-to-v6

Mixed means the address object contains both IPv4 and IPv6 addresses. I believe the broken cases are:
mixed-to-v6
mixed-to-v4
mixed-to-mixed
v4-to-mixed
v4-to-v4
v4-to-v6

Might be fallout from: https://github.com/google/capirca/issues/129

rarcotvmw:
I do see an issue with the mixed filter.
For mixed filter, the generator generates both IPV4 and IPV6 addresses for source and destination and there are no validations for some of the cases mentioned. Will work on the fix.

ankenyr:
Can we resolve this now @SecurityForDotes ?

SecurityForDotes:
Yup.

PR #138, opened 2017-07-06
Nsxv generator - Fixed generating invalid mix of IPv6 <-> IPv4 addres…

rarcotvmw:
…ses in the output for mixed filter

Issue # https://github.com/google/capirca/issues/137

Mixed filter generates both Ipv4 and IPv6 address for source and destination.
There was no check to see if source and destination addresses had a mismatch.
Added code to drop the mismatch IP if source and destination address are not valid for mixed filter

Eg
IPv4 to IPv4 +IPv6 (results in IPv6 being dropped from term and vice versa)
IPv6 to IPv6 +IPv4 (results in IPv4 being dropped from term and vice versa)

Tests
Changed the Mock method to self.naming.GetNetAddr.side_effect instead of Mock.self.naming.GetNetAddr('NTP_SERVERS') as it was not returning the ip address values and some cases were not checked.

Testing Done:
Tested manually by running the script and creating section in NSX Manager.
Added test cases

$ python -m unittest discover -s . -p '*_test.py'
Ran 597 tests in 84.372s

OK

### DIFF
--- a/lib/nsxv.py
+++ b/lib/nsxv.py
@@ -209,6 +209,11 @@ class Term(aclgenerator.Term):
     source_addr = []
     destination_addr = []
 
+    source_ipv4_address = False
+    source_ipv6_address = False
+    dest_ipv4_address = False
+    dest_ipv6_address = False
+
     for af in af_list:
       # source address
       if self.term.source_address:
@@ -219,14 +224,13 @@ class Term(aclgenerator.Term):
           source_address = nacaddr.ExcludeAddrs(
               source_address,
               source_address_exclude)
-        if not source_address:
-          logging.warn(self.NO_AF_LOG_ADDR.substitute(term=self.term.name,
-                                                      direction='source',
-                                                      af=self.filter_type))
-        if not source_addr:
-          source_addr.extend(source_address)
-        else:
-          source_addr = source_address
+
+        if source_address:
+          if af == 4:
+            source_ipv4_address = True
+          else:
+            source_ipv6_address = True
+        source_addr.extend(source_address)
 
       # destination address
       if self.term.destination_address:
@@ -238,11 +242,28 @@ class Term(aclgenerator.Term):
           destination_address = nacaddr.ExcludeAddrs(
               destination_address,
               destination_address_exclude)
-        if not destination_address:
-          logging.warn(self.NO_AF_LOG_ADDR.substitute(term=self.term.name,
-                                                      direction='destination',
-                                                      af=self.filter_type))
+
+        if destination_address:
+          if af == 4:
+            dest_ipv4_address = True
+          else:
+            dest_ipv6_address = True
         destination_addr.extend(destination_address)
+
+    # Drop the mismatch IP if source and destination address are not valid
+    # for mixed filter
+    if self.filter_type == 'mixed':
+      if source_addr and destination_addr:
+        if source_ipv4_address and not dest_ipv4_address:
+          source_addr = self._DropMismatchIpAddress(source_addr, 'source', 4)
+        elif source_ipv6_address and not dest_ipv6_address:
+          source_addr = self._DropMismatchIpAddress(source_addr, 'source', 6)
+        elif dest_ipv4_address and not source_ipv4_address:
+          destination_addr = self._DropMismatchIpAddress(destination_addr,
+                                                 'destination', 4)
+        elif dest_ipv6_address and not source_ipv6_address:
+          destination_addr = self._DropMismatchIpAddress(destination_addr,
+                                                 'destination', 6)
 
     # ports
     source_port = None
@@ -352,6 +373,18 @@ class Term(aclgenerator.Term):
     stripped_ret_lines = [re.sub(r'\s+', ' ', x).rstrip() for x in ret_lines]
     ret_str.extend(stripped_ret_lines)
     return ''.join(ret_str)
+
+  def _DropMismatchIpAddress(self, ipaddress, direction_address, af):
+    modified_ipaddress = []
+    for addr in ipaddress:
+      if ((af == 4 and not type(addr) is nacaddr.IPv4)
+          or (af == 6 and not type(addr) is nacaddr.IPv6)):
+        modified_ipaddress.append(addr)
+      else:
+        logging.warn('Term %s has IPv4/IPv6 mismatch for %s for mixed '
+                     'address family, dropping %s ipaddress.', self.term.name,
+                     direction_address, addr)
+    return modified_ipaddress
 
   def _ServiceToString(self, proto, sports, dports, icmp_types):
     """Converts service to string.
--- a/lib/nsxv.py
+++ b/lib/nsxv.py
@@ -209,10 +209,11 @@ class Term(aclgenerator.Term):
     source_addr = []
     destination_addr = []
 
-    source_ipv4_address = False
-    source_ipv6_address = False
-    dest_ipv4_address = False
-    dest_ipv6_address = False
+    source_v4_addr = []
+    source_v6_addr = []
+    dest_v4_addr = []
+    dest_v6_addr = []
+
 
     for af in af_list:
       # source address
@@ -227,10 +228,10 @@ class Term(aclgenerator.Term):
 
         if source_address:
           if af == 4:
-            source_ipv4_address = True
+            source_v4_addr = source_address
           else:
-            source_ipv6_address = True
-        source_addr.extend(source_address)
+            source_v6_addr = source_address
+        source_addr = source_v4_addr + source_v6_addr
 
       # destination address
       if self.term.destination_address:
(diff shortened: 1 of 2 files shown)


lib/nsxv.py
```diff
@@ -353,6 +374,18 @@ def __str__(self):
     ret_str.extend(stripped_ret_lines)
     return ''.join(ret_str)
 
+  def _DropMismatchIpAddress(self, ipaddress, direction_address, af):
+    modified_ipaddress = []
+    for addr in ipaddress:
+      if ((af == 4 and not type(addr) is nacaddr.IPv4)
```
ankenyr: Do not compare with type, use isinstance.
https://stackoverflow.com/questions/30049410/why-use-isinstance-instead-of-type

lib/nsxv.py
```diff
@@ -353,6 +374,18 @@ def __str__(self):
     ret_str.extend(stripped_ret_lines)
     return ''.join(ret_str)
 
+  def _DropMismatchIpAddress(self, ipaddress, direction_address, af):
```
ankenyr: Please add a docstring to this function if you intend to keep it. Look at other functions for examples. It should have a sentence description saying what it does and a section explianing arguments and returns and raises

ankenyr:
Made a couple of quick comments and assigned this to @SecurityForDotes for more review. I think there is a much easier way to do this. You basically want to drop IPs when it is not mixed right? I think a simple list comprehension would bring this down to only a single line for source and destination. It is late so I will give this further review with @SecurityForDotes in the morning.

rarcotvmw:
Will wait for review comments from @SecurityForDotes also and make the changes accordingly.

lib/nsxv.py
```diff
@@ -209,6 +209,11 @@ def __str__(self):
     source_addr = []
     destination_addr = []
 
+    source_ipv4_address = False
+    source_ipv6_address = False
+    dest_ipv4_address = False
+    dest_ipv6_address = False
```
SecurityForDotes: So I think this code can be made a bit cleaner. Instead of defining these boolean that are used for the flow control at line 256, you could instead create two more lists and rename the existing lists.
source_v4_addr
source_v6_addr
dest_v4_addr
dest_v6_addr

 In addition an optimization could occur that if after you finished source section, if you only v4 were used, you can skip the v6 for dest loop since they would be dropped anyways, but that's totally optional and not sure which is more performant.

And in the loop on 217 where it goes through each af_list, just use the correct list for the storage of the values.

Following that where the flow control starts at 256, you can just the size or status of the lists created as an implicit way to determine if their was v6 address or not. Then you can just copy the values to the final array, or just join the arrays together.

SecurityForDotes: Another option would be if there is a way to guarantee that naming always returns a sorted list of address. You could compare the first and last values and if they are the same AF family then it only contains one family. This would definitely simplify the process. I took a look at naming and it does appear to sort the returned result but haven't tested if it actually makes it way back through policyreader, then policy then generators to confirm.

lib/nsxv.py
```diff
@@ -238,12 +242,29 @@ def __str__(self):
...
+        if source_ipv4_address and not dest_ipv4_address:
+          source_addr = self._DropMismatchIpAddress(source_addr, 'source', 4)
+        elif source_ipv6_address and not dest_ipv6_address:
+          source_addr = self._DropMismatchIpAddress(source_addr, 'source', 6)
+        elif dest_ipv4_address and not source_ipv4_address:
+          destination_addr = self._DropMismatchIpAddress(destination_addr,
+                                                 'destination', 4)
+        elif dest_ipv6_address and not source_ipv6_address:
+          destination_addr = self._DropMismatchIpAddress(destination_addr,
+                                                 'destination', 6)
```
SecurityForDotes: One use case that I'm not sure if the correct behavior happened is:

v4-to-v6 (and vice versa)
it currently drops the source address and keeps the destination address. I'm not sure what the best behavior would be. It feels like the generator gets to a situation where it removes all the address for a side and thus making it that side any, that a warning or maybe even a dropping of the whole term is warranted. But i'm not sure. What is your take on it?
@ankenyr  what do you think should happen in this situation (lets say from an ideal perspective even if it's not what other generators do)

tests/lib/nsxv_test.py
```diff
@@ -234,6 +234,40 @@
   }
   """
 
+SOURCE_IPv6_MISMATCH = """\
```
SecurityForDotes: I think naming these a bit longer is a good idea.

I've been doing:
mixed-to-v6
mixed-to-v4
mixed-to-mixed
v4-to-mixed
v4-to-v4
v4-to-v6
v6-to-v6

Something like that that clarifies the intent of what it is doing.

lib/nsxv.py
ankenyr: The whole idea of this was to eliminate rules which are not valid, a rule that is src ipv4 and dst to ipv6 is invalid and thus should be removed. If the result after filtering results in 0 addresses on one side the expectation is not to add in an any, that would open up a  huge hole. It should remove the rule in this case.

SecurityForDotes:
I'll take a look at tomorrow. Sorry for the delay.

tests/lib/nsxv_test.py
```diff
@@ -747,108 +794,156 @@ def testBadHeaderCase4(self):
...
+
+  def testV4ToV6(self):
+    self.naming.GetNetAddr.side_effect = [
+        [nacaddr.IP('10.0.0.0/8'), nacaddr.IP('172.16.0.0/12'),
+         nacaddr.IP('192.168.0.0/16')],
         [nacaddr.IP('2001:4860:8000::/33')]]
-    pol = policy.ParsePolicy(MIXED_HEADER + DEST_IPv4_MISMATCH, self.naming,
-                             False)
+
+    source_addr, dest_addr = self.getSourceDestAddresses(MIXED_HEADER +
```
SecurityForDotes: This shouldn't return a term because it drops both sides (and it correctly does). This test is relying on the return is 0 based on the search of the xml. Perhaps add a comment to make it explicit that the expectation is that no term gets generated. 

In fact I think it might be useful to have an explicit check that no term was created, however it's up to you I won't require it.

tests/lib/nsxv_test.py
```diff
@@ -747,108 +794,156 @@ def testBadHeaderCase4(self):
...
+    self.naming.GetNetAddr.assert_has_calls(
+        [mock.call('INTERNAL'), mock.call('SOME_HOST')])
+
+  def testV6ToV4(self):
+    self.naming.GetNetAddr.side_effect = [
+        [nacaddr.IP('2001:4860:8000::/33')],
+        [nacaddr.IP('10.0.0.0/8'), nacaddr.IP('172.16.0.0/12'),
+         nacaddr.IP('192.168.0.0/16')]]
+
+    source_addr, dest_addr = self.getSourceDestAddresses(MIXED_HEADER +
```
SecurityForDotes: Same comment as testV4ToV6

tests/lib/nsxv_test.py
```diff
@@ -861,15 +956,16 @@ def testNsxvStrForDestinationIpv4Mismatch(self):
...
-      addr_type = dest_addr.find('type').text
-      self.assertEquals(addr_type, 'Ipv6Address')
+    source_addr = root.findall('./rule/sources/source')
+    dest_addr = root.findall('./rule/destinations/destination')
 
-    self.naming.GetNetAddr.assert_has_calls(
-        [mock.call('SOME_HOST'), mock.call('GOOGLE_DNS')])
+    return source_addr, dest_addr
+
+  def verifyAddressType(self, address, type):
```
SecurityForDotes: What if you make type a list, and do assert in, that way for the two test cases where it could be either you don't need to replicate the code?

tests/lib/nsxv_test.py
```diff
@@ -234,40 +234,87 @@
...
 
+MIXED_TO_MIXED = """\
+  term mixed_to_mixed {
+    source-address:: GOOGLE_DNS
+    destination-address:: GOOGLE_DNS
+    action:: accept
+  }
+  """
+
+MIXED_TO_ANY = """\
```
SecurityForDotes: please add an additional test that covers ANY to Mixed.

tests/lib/nsxv_test.py
rarcotvmw: Done

tests/lib/nsxv_test.py
rarcotvmw: Done

tests/lib/nsxv_test.py
rarcotvmw: Added the test case.

tests/lib/nsxv_test.py
rarcotvmw: Not clear about this change. How will making type a list help?  Can you please let me know?
Or else this method should take the source addr, destination add, type and loop through each source addr and destination addr and check the type.

tests/lib/nsxv_test.py
```diff
@@ -747,108 +794,156 @@ def testBadHeaderCase4(self):
...
-    self.assertIsNotNone(source_address)
-    for source_addr in source_address:
-      addr_type = source_addr.find('type').text
-      self.assertEquals(addr_type, 'Ipv6Address')
+    source_addr, dest_addr = self.getSourceDestAddresses(MIXED_HEADER +
+                                                         MIXED_TO_MIXED)
+    self.assertFalse(len(source_addr) == 0)
+    for saddr in source_addr:
+      saddr_type = saddr.find('type').text
+      self.assertIn(saddr_type, ('Ipv4Address', 'Ipv6Address'))
```
SecurityForDotes: So I think we might have a problem here. Is this ever not going to be true? Its always going to be one or the other right (unless the term is missing). This doesn't check that both exist which I think is the true point of this test.

tests/lib/nsxv_test.py
SecurityForDotes: Let me try to expand on this a bit.

in an earlier section (which I commented on (testMixedToMixed and testMixedToAny)), this function was duplicated in line, because the assertEquals was replaced with "self.assertIn(saddr_type, ('Ipv4Address', 'Ipv6Address'))".

my comment originally was, if this function is changed to make type a list instead of a string, the function could be updated to be assertIn, that way that function can be used for all the tests.

However in my comment that I made today earlier, that assertIn doesn't test exactly what we need (because it doesn't ensure both are true), if that can be solved and the refactor still works I think it should be done.

Hopefully that helps explain my intention a bit better.

tests/lib/nsxv_test.py
rarcotvmw: Yes. I think we need to explicitly check if V4 and V6 are present

rarcotvmw: Made the changes.

Issue #121, opened 2017-02-08
object-group support for ios


sjtarik:
I saw the following short-comings for object group support for IOS. 

1. Are object-groups with inet6 supported? What I see is one or the other.
2. ICMP and TCP options are not rendered for object-groups (e.g., tcp-established). 
3. Protocol numbers are always integers which break NXOS like systems if portgroups are present. _PROTO_INT flag should be honored.

sjtarik:
Multiple instances of same object-group with different ip sets will be created, result in unintended output if multiple subnets passed as source and destination.

```
header {
  comment:: "first-group-acl-a"
  target:: cisco first-group-acl-a object-group
}
term permit-all-from-groupa {
  source-address:: GROUPA
  action:: accept
}
term permit-all-from-groupa {
  destination-address:: GROUPB
  action:: accept
}

header {
  comment:: "first-group-acl-b"
  target:: cisco first-group-acl-b object-group
}
term permit-all-from-groupb {
  source-address:: GROUPA GROUPB <========== All addresses will be consolidated under A
  action:: accept
}



object-group network ipv4 GROUPA
 1.1.1.1/32
 2.2.2.2/32
 3.3.3.3/32
 4.4.4.4/32
exit



object-group network ipv4 GROUPA
 1.1.1.1/32
 2.2.2.2/32
 3.3.3.3/32
exit

object-group network ipv4 GROUPB
 1.1.1.1/32
 4.4.4.4/32
exit

! $Id:$
! $Date:$
! $Revision:$
no ip access-list extended first-group-acl-a
ip access-list extended first-group-acl-a
 remark $Id:$
 remark first-group-acl-a


 remark permit-all-from-groupa
 permit ip net-group GROUPA any


 remark permit-all-from-groupa
 permit ip any net-group GROUPB

exit

no ip access-list extended first-group-acl-b
ip access-list extended first-group-acl-b
 remark $Id:$
 remark first-group-acl-b


 remark permit-all-from-groupb
 permit ip net-group GROUPA any

exit

```

sjtarik:
Final issue is the hardest one to crack. When optimization is done, group context is lost if any subnet under the main group is optimized/merged. In below example SUBGROUP1 and SUBGROUP2 is merged so that their parent token is SUBGROUP1 and no longer GROUPA or GROUPB. Since cisco generator use the parent token to name object group, both object-group names become SUBGROUP1 when two filters are generated.

GROUPA = SUBGROUP1
                   SUBGROUP2
[...]

Issue #70, opened 2016-06-09
Potential global state trouble with policy.py's _OPTIMIZE flag


jzohrab:
policy.py has the following in TranslateTerms():

```
    for term in terms:
      ...
      term.AddressCleanup(_OPTIMIZE)
      # Reset _OPTIMIZE global to default value
      globals()['_OPTIMIZE'] = True
      term.SanityCheck()
      term.translated = True
```

_OPTIMIZE could be set to False globally at the start of an aclgen run, but when TranslateTerms is called with a policy with multiple terms, it will be False for the first term, and then True for all remaining terms.

I don't have a test case illustrating this behaviour yet, but it shouldn't be hard to come up with one.

I don't know if the global _OPTIMIZE setting will influence multiple policies.  I'll need to look further into what the implications for this are.

At any rate, global state is bad, and the current changing of global state during the run feels incorrect.

jzohrab:
Same issue with _SHADE_CHECK, which is set at a global level but then potentially overridden with a call in policy.py:

```
  def _DetectShading(self, terms):
    ...
    globals()['_SHADE_CHECK'] = False
```

Because of this global state, the test `python -m unittest tests.lib.policy_test.PolicyTest.testTermEquality` was passing when it was run on its own, but failing when run as part of the full test suite.  At least, that's my initial hypothesis, I didn't have time to investigate fully.

GoodEveningMiss:
seconded, saw this with _OPTIMIZE and _SHADE_CHECK

sjtarik:
This is still an issue. Just for consistent behavior, I suggest setting FLAGS.optimize default value to True and removing the statement pointed out by @jzohrab.

ankenyr:
This should be fixed as of a few pushes ago

PR #141, opened 2017-07-26
fixing the cisco object-group issue when optimization is on and paren…

sjtarik:
This is to fix issue #121 and also partially address the #70  for optimize flag. Test file changes are mostly due to #70 since current behavior is honoring the default value of optimize=False for first term but for the following term overwriting that value with True and optimize the ip address lists with merging any subnets possible.


ankenyr:
Looks like the nacaddr change caused a problem.

Seems that if you have two terms like the following
term accept-something {
  destination-address:: FOO
  protocol:: tcp
  action:: accept
}

term accept-another {
  destination-address:: BAR
  destination-address:: FOO
  protocol:: udp
  action:: accept
}
It seems that in the addressbook BAR would not be present in the IP addresses.
We have had this problem before in the SRX platform and it may be worth looking there for possible solutions.

sjtarik:
@ankenyr Just to clarify this happens when FOO and BAR can be summarized, right? Otherwise this does not happen I assume.

sjtarik:
This issue is independent of the recent changes. I was able to recreate the issue without my changes as well. Updated the [issue](https://github.com/google/capirca/issues/121) report with examples.

PR #142, opened 2017-07-26
Fix for Issue 77

pw378:


ankenyr:
That error is due to the integration tests, I fixed those, if you sync to head it should be fixed.

aclgen.py
```diff
@@ -24,7 +24,9 @@
 from __future__ import unicode_literals
 
 __author__ = 'pmoody@google.com'
+__author__ = 'watson@google.com'
```
ankenyr: author tags are no longer in style Tony ;) I have been considering removing all the existing ones from the code actually. I will share with you an internal document talking about this.

pw378: ack

PR #144, opened 2017-08-08
srx duplicate policy check (for same from/to zone pair)

sjtarik:
…m-zone to-zone pair in the same policy file which would overwrite each other on srx platform


lib/junipersrx.py
```diff
@@ -318,6 +322,7 @@ def _TranslatePolicy(self, pol, exp_info):
     Raises:
       UnsupportedFilterError: An unsupported filter was specified
       UnsupportedHeader: A header option exists that is not understood/usable
+      SRXDuplicatePolicy: A policy with multiple definition for same from-zone to-zone pair
```
ankenyr: nit: A policy with duplicate from-zone to-zone pairs.

ankenyr:
Going to need to roll this back, seems this checks for duplicate names across the boundaries which it should not. It is common to use include files to abstract certain terms that need to be used everwhere. Those include files would be used in multiple headers and that would trigger this error message.

sjtarik:
@ankenyr, could you give an example of boundary condition? Or even better point to a unit test. Let me fix it since I need this logic to resolve conflicts in address book resolution?

PR #145, opened 2017-08-09
fix proposal for conflicting srx address-sets and ios object-group definitions due to optimization and parent_token usage

sjtarik:
without a no statement to remove object-groups, changes to object groups will be merged. tested on nxos and ios-xe/catos versions.


sjtarik:
@ankenyr Rob, could you review this implementation for ios? Basically it keeps a dict of already used object groups and their content (list of IP addresses). In case of conflict, same object group name but with a different list of IPs, it appends _{\d} to the end of object group name derived from the parent token of first IP address in the list of addresses. If this is not too hacky, I can add more unit tests and revise the pull request. Thanks.

sjtarik:
@ankenyr Rob adding the iosxe (cat4500x, cat3850 alike, cat6k) support for ios object-groups. I dismissed the port-group support for cat4k since it is not aligned with capirca's regular handling of port groups. Also earlier there is a bug where tcp and icmp options are silently ignored for object-groups. This commit also fixes this problem.

ankenyr:
Going to mark this one as closed since it is pretty large.
075ffa40a8c6daa086d528a6ae6c727e4696acdc has the mechanism to preserve tokens for platforms that use addressbooks. If there are any that have been missed lets fix those in follow up PRs.

PR #151, opened 2018-05-02
convert plain text docs to Markdown

russStarr:


russStarr:
`I signed it!`

ankenyr:
Hey @russStarr,
The docs here are actually not fully up to date. We tend to rely on the wiki pages linked at the top.
(https://github.com/google/capirca/wiki/)
I am unsure if there is a way to keep those synced and think it might be worth just removing the docs files so that everything is in one place. Is there a reason that keeping the docs folder would be helpful to you over the wiki?

russStarr:
I should have known to check the wiki link since that's where I tend to place docs on github enterprise. I suppose you can ignore the PR and I'll go look at the wiki.

PR #152, opened 2018-05-14
Fragments support in Cisco

oribit:
Cisco ACL allows to handle specifically fragments adding the keyword "fragments" to the end of any L3 ACE. This PR will allow that through the use of "fragments" as an option in Capirca policy.

### DIFF
--- a/capirca/lib/cisco.py
+++ b/capirca/lib/cisco.py
@@ -401,6 +401,9 @@ class Term(aclgenerator.Term):
         and ('tcp-established' in opts or 'established' in opts)):
       if 'established' not in self.options:
         self.options.append('established')
+    if ('ip' in protocol) and ('fragments' in opts):
+      if 'fragments' not in self.options:
+          self.options.append('fragments')
 
     # ports
     source_port = [()]
@@ -632,7 +635,9 @@ class Term(aclgenerator.Term):
   def _FixOptions(self, proto, option):
     """Returns a set of options suitable for the given protocol.
 
-    In practice this is only used to filter out 'established' for UDP.
+    Fix done:
+    - Filter out 'established' for UDP.
+    - Filter out 'fragments' for TCP/UDP
 
     Args:
       proto: str or int, protocol
@@ -645,6 +650,12 @@ class Term(aclgenerator.Term):
     if ((proto == self.PROTO_MAP['udp'] or proto == 'udp')
         and 'established' in sane_options):
       sane_options.remove('established')
+    if ((proto == self.PROTO_MAP['tcp'] or proto == 'tcp')
+        and 'fragments' in sane_options):
+      sane_options.remove('fragments')
+    if ((proto == self.PROTO_MAP['udp'] or proto == 'udp')
+        and 'fragments' in sane_options):
+      sane_options.remove('fragments')
     return sane_options
 
   def _TermletToStr(self, action, proto, saddr, sport, daddr, dport,
@@ -831,7 +842,8 @@ class Cisco(aclgenerator.ACLGenerator):
                          'owner'}
 
     supported_sub_tokens.update({'option': {'established',
-                                            'tcp-established'},
+                                            'tcp-established',
+                                            'fragments'},
                                  # Warning, some of these are mapped
                                  # differently. See _ACTION_TABLE
                                  'action': {'accept', 'deny', 'reject', 'next',
--- a/capirca/lib/cisco.py
+++ b/capirca/lib/cisco.py
@@ -650,12 +650,6 @@ class Term(aclgenerator.Term):
     if ((proto == self.PROTO_MAP['udp'] or proto == 'udp')
         and 'established' in sane_options):
       sane_options.remove('established')
-    if ((proto == self.PROTO_MAP['tcp'] or proto == 'tcp')
-        and 'fragments' in sane_options):
-      sane_options.remove('fragments')
-    if ((proto == self.PROTO_MAP['udp'] or proto == 'udp')
-        and 'fragments' in sane_options):
-      sane_options.remove('fragments')
     return sane_options
 
   def _TermletToStr(self, action, proto, saddr, sport, daddr, dport,
--- a/tests/lib/arista_test.py
+++ b/tests/lib/arista_test.py
@@ -171,7 +171,8 @@ SUPPORTED_SUB_TOKENS = {
         'version-2-multicast-listener-report',
     },
     'option': {'established',
-               'tcp-established'}
+               'tcp-established',
+               'fragments'}
 }
 
 # Print a info message when a term is set to expire in that many weeks.
--- a/tests/lib/cisco_test.py
+++ b/tests/lib/cisco_test.py
@@ -284,6 +284,14 @@ term good_term_19 {
   action:: accept
 }
 """
+GOOD_TERM_20 = """
+term good_term_20 {
+  source-address:: SOME_HOST
+  destination-address:: SOME_HOST
+  option:: fragments
+  action:: accept
+}
+"""
 LONG_COMMENT_TERM = """
 term long-comment-term {
   comment:: "%s "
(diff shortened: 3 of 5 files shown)


ankenyr:
Hey orbit. Will be out of the office for the next two weeks but will look
at this when I get back.

ankenyr:
Hey Orbit, sorry for letting these linger for so long. Looking at this now

lib/cisco.py
```diff
@@ -827,7 +838,8 @@ def _BuildTokens(self):
                          'owner'}
 
     supported_sub_tokens.update({'option': {'established',
-                                            'tcp-established'},
+                                            'tcp-established',
+                                            'fragments'},
```
ankenyr: I think what you are looking for is 'is-fragment' not 'fragments'
https://github.com/google/capirca/blob/24568835dd916652824cf55f25cc66a3a16e716a/lib/aclgenerator.py#L361

oribit: This is one of the things that makes Cisco so "lovable"... "is-fragment" is a keyword to be used with "fragment-type", but when you want to make an ACE exclusively specific for fragments, you just need to add "fragments" keyword to it, and then it will be evaluate only when fragment offset is not 0. This PR is to support this second "fragments".

tests/lib/arista_test.py
```diff
@@ -171,7 +171,8 @@
         'version-2-multicast-listener-report',
     },
     'option': {'established',
-               'tcp-established'}
+               'tcp-established',
```
ankenyr: Do these other platforms actually support this?

tests/lib/cisco_test.py
```diff
@@ -798,5 +807,15 @@ def testProtoInts(self):
     self.failUnless('permit udp any any range 1024 65535' in str(acl),
                     str(acl))
 
+  def testFragments(self):
+    self.naming.GetNetAddr.return_value = [nacaddr.IP('10.0.0.0/24')]
+    acl = cisco.Cisco(policy.ParsePolicy(GOOD_HEADER + GOOD_TERM_20,
+                                         self.naming), EXP_INFO)
+    first_string = 'permit ip 10.0.0.0 0.0.0.255 10.0.0.0 0.0.0.255 fragments'
```
ankenyr: instead of first_string lets rename this to "expected"

tests/lib/arista_test.py
oribit: I really don't know, but the test fails because it has Cisco dependencies.

tests/lib/cisco_test.py
oribit: changed

oribit:
@zombah Thank you very much for that link. Indeed, I can see there the support for fragments is the same than in the Cisco ACL, both of them check for the FO bit set. 

@ankenyr It seems that the documentation from Arista explains that it's supported. Having said that, Capirca support for Arista (and Brocade) is just an "import Cisco", so whatever is added there it will be added automatically to Arista and Brocade. This is a limitation in the way of these two brands are supported in Capirca, because if anything is changed in Cisco, it will be included. Here the problem was with the token in the tests exclusively, those were the file changed, but it doesn't really matter if the test fails or not, because it's going to be included anyway.

tests/lib/cisco_test.py
```diff
@@ -364,7 +372,8 @@
         'version-2-multicast-listener-report',
     },
     'option': {'established',
-               'tcp-established'}
+               'tcp-established',
+               'fragments'}
```
ankenyr: There is already an option called is-fragment. For the sake of keeping things unified can we use that?

oribit: Although it has the same functionality, using if-fragment add more complexity to an ACL. The option "is-fragment" has to be used together with "fragment-type". This should be used when an ACL has to be configured with a deep granularity. To handle only fragments, or to be more specific non-initial fragments, the keyword "fragments" added to any ACE, makes that ACE only applicable to a non-initial fragment. Cisco provides both options, and I think Capirca could support both.

oribit: You can see here a good explanation about how Cisco implement fragments support: https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5xx/ipaddress/63x/b-ip-addresses-cg-63x-ncs5xx/b-ip-addresses-cg-63x-ncs5xx_chapter_01.pdf 

- Page 17: the use of "fragments" keyword
- Page 19: use case for fragment-type and is-fragment

capirca/lib/cisco.py
```diff
@@ -398,6 +398,9 @@ def __str__(self):
         and ('tcp-established' in opts or 'established' in opts)):
       if 'established' not in self.options:
         self.options.append('established')
+    if ('ip' in protocol) and ('fragments' in opts):
```
ankenyr: I see you are checking on if 'ip' is in protocol here. Is this needed? Other protocols can be fragmented, yes?

oribit: If I'm not remembering wrong (and also reviewing the code), indeed, that 'if' is making sure that fragments are only used with 'ip' protocol. Maybe there are other protocols supporting fragmentation (or something similar), but this is made here to avoid a common confusion of using fragmentation with TCP or UDP protocols.

ankenyr: This likely won't work for other platforms, ciscoxr relies on ipv4 being in the protocol. What confusion are you trying to avoid with TCP and UDP with regards to fragments? Is this that you cannot use layer 4 information in the rule when you have fragments enabled?

If so I think the usage of ip and ipv6 in this area is a bit confusing considering the project as a whole.
https://github.com/google/capirca/blob/master/capirca/lib/aclgenerator.py#L92
https://github.com/google/capirca/blob/master/capirca/lib/cisco.py#L344

The use of protocol here (in cisco.py in general, not your CL specifically) is incorrect I believe. I ran into other issues with regards to how protocols are used in Capirca in a few CLs. So maybe this should be rethought.

capirca/lib/cisco.py
```diff
@@ -640,6 +645,12 @@ def _FixOptions(self, proto, option):
     if ((proto == self.PROTO_MAP['udp'] or proto == 'udp')
         and 'established' in sane_options):
       sane_options.remove('established')
+    if ((proto == self.PROTO_MAP['tcp'] or proto == 'tcp')
```
ankenyr: I might be incorrect here as I have not worked much on this section of the code but I believe this is used to prevent the printing of the word established in cases where a udp established term would be used. The established field is useful in some cases. If fragments are not valid to be used with UDP or TCP it is likely that instead of filtering we should be marking these invalid by erroring out with a message.

capirca/lib/cisco.py
oribit: Sorry, this was submitted some time ago and I didn't remember clearly. But, reviewing the code, this part is just to add the "fragments" keyword. "Term.options" should contain any option defined in the policy, like "log", so this is the way to add "fragments". This control wouldn't be necessary if when we reach this point, we are 100% sure that options are set correctly, so for example, fragments are not used with other protocol than "ip", but I think we can't be 100% sure, so it's better double check that the ACL is going to be built it makes sense.

capirca/lib/cisco.py
oribit: This whole method "_FixOptions", from my point of view it doesn't make sense. If an option is invalid, I totally agree with you that it should be pointed instead of just "fixing under the hood". Indeed, imho this is a bug. If someone use the tcp-established option with udp protocol, capirca shouldn't create the ACL. For Cisco.py, there were only two valid sub-tokens for token option: "tcp-established" and "established". The decision done when this was implemented was to avoid checking this options but fixing any wrong option used, so I followed the same principle. I've just changed that and removed that part. But then, if an option is wrongly used, this option is removed without any notification (this is how Capirca works). I could create a new PR to improve this.

ankenyr: sgtm

PR #157, opened 2018-08-14
Fix header comment generation

zombah:
With strict condition for platfrom
cisco and filter type standard

Signed-off-by: Misha Komarovskiy <zombah@gmail.com>

### DIFF
--- a/capirca/lib/cisco.py
+++ b/capirca/lib/cisco.py
@@ -999,7 +999,11 @@ class Cisco(aclgenerator.ACLGenerator):
         # add a header comment if one exists
         for comment in header.comment:
           for line in comment.split('\n'):
-            target.append(' remark %s' % line)
+            if (self._PLATFORM == 'cisco' and filter_type == 'standard' and
+                filter_name.isdigit()):
+              target.append(' access-list %s remark %s' % (filter_name, line))
+            else:
+              target.append(' remark %s' % line)
 
         # now add the terms
         for term in terms:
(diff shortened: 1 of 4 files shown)


ankenyr:
Can you create tests for these?

zombah:
@ankenyr You mean to expand testRemark with header remark check? 
I think i can try, but cisco_test.py already failing for me at current master with FAIL: testExpiringTerm (tests.lib.cisco_test.CiscoTest)  is it know problem?

ankenyr:
I am looking for tests to be made for cisco to confirm it has the expected results and that its child variants have their expected results.
As for tests failing that is unexpected, I don't see travis reporting it is failing. Got any more details you can share on what exactly is happening?

zombah:
That was temporary fail, after couple runs it vanished, i didn't figured why.
I expanded cisco testRemark for standard filter type.

zombah:
I don't have access to arista platform so i'm not sure of proper remarking style there, but i can forge remark test based on their documentation.

zombah:
@ankenyr @Manningham I added tests, please review

zombah:
@ankenyr it seems my pr need rebase now, do i need to care about this or there is no chances to receive any reviewson this patch?

ankenyr:
So sorry I let this linger for so long. I changed things to resolve the conflict and fix it for the latest addition of verbosity settings. This should look good, just letting travis do its thing and will accept it after.

Issue #166, opened 2018-12-11
TermPortToProtocol converts some port numbers to protocol names not supported on Arista


amybabay:
Specifically, when I created a policy that allowed TCP on port 5631, the port number got converted to 'pcanywhere-data'. That wasn't recognized by my Arista (tested on EOS v 4.19.1F and 4.21.1.1F), preventing the generated ACL from being applied.

The same issue exists with UDP port 5632 converting to 'pcanywhere-status', and I think a few others as well.

The simplest option would be to not try to convert the ports at all and leave them as literal port numbers -- deleting the lines that call TermPortToProtocol gives me exactly what I want.
But do other people have uses where it's important to use the names instead of port numbers? If so, would making it optional be worthwhile? Or is the best approach for now to just fix specific mappings that don't work?

ankenyr:
I think this is a repercussion of abstracting this behavior from the individual generators into the cisco generator. This mapping will likely work on an asa as looking back in time that port name mapping was in that generator.

https://github.com/google/capirca/commit/9c08416e06dc26e6af153b48f14bc4386316bad3

Preserving the names would be good but creating logic to only print out names that are supported by the generator would be wanted. Let me know if you wish to take this on and provide a PR.

amybabay:
Got it, that makes sense! Yes, I can work on that and provide a PR.

PR #169, opened 2019-01-16
Addresses issue #166: Separating Cisco port/service name mappings from Arista ones

amybabay:
Specifically, replaced TermPortToProtocol (which had Cisco mappings
defined inside it) with PortMap class that defines both sets of mappings
and provides function to look up service name based on port + platform


ankenyr:
Thanks for this one also.

PR #173, opened 2019-05-28
fortigate firewall policy support

Ali-aqrabawi:
in this pull request i added support for `fortigate` policies. 

example `.pol` 

    header {
      target:: fortigate from-id 2
      target:: paloalto from-zone internal to-zone external
      target:: juniper edge-filter
      target:: cisco edge-filter
    }

    term allow-https-web {
      destination-address:: WEB_SERVERS
      destination-port:: WEB_SERVICES
      protocol:: tcp
      action:: accept
    }

    term customers-policy {
      source-address:: CUSTOMERS
      destination-address:: GOOGLE_PUBLIC_DNS_ANYCAST
      destination-port:: WEB_SERVICES
      protocol:: tcp
      action:: deny
    }

    term customers-policy2 {
      #source-interface:: wan1
      #destination-interface:: wan2
      source-address:: ALI_HOME
      destination-address:: CRM
      destination-port:: WEB_SERVICES
      protocol:: tcp
      action:: deny
      logging:: true
    }

    term deny-any-any {
      action:: deny

    }

    term deny-icmp {
      action:: deny
      protocol:: icmp

    }
    term accept-whois {
      protocol:: udp
      destination-port:: WHOIS
      action:: accept
    }


capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+      raise FortiGateValueError('%r protocol is not supported by Fortigate, supported protocols = %r' % (
+        protocol, FortigatePortMap._PROTO_MAP.keys()))
+
+    if isinstance(f_proto, str):
+      return f_proto
+    elif port:
+      return f_proto[port]
+
+    else:
+      raise FortiGateFindServiceError('failed to get service from %r protocol and %r port' % (protocol, port))
```
ankenyr: Same here, please limit to 80 characters. You can break this up by following this as an example.
https://github.com/google/capirca/blob/b4d20fe2ee968d9b44fe920399461ef94c5df7ea/capirca/lib/junipersrx.py#L375

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class Term(aclgenerator.Term):
```
ankenyr: Add docstring please

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+                                    '\tnext'])
+    type(self)._FW_DUP_CHECK.add(addr)
+
+  def _add_service_to_fw_services(self, protocol, service):
+    if service in type(self)._FW_DUP_CHECK:
+      return
+
+    type(self).FW_SERVICES.extend(['\tedit %s' % service,
+                                   '\t\tset protocol TCP/UDP',
+                                   '\t\tset %s-portrange %s' % (protocol.lower(), service),
```
ankenyr: Line too long, you can fix this by moving the list to its own new line and indenting 
```
    type(self).FW_SERVICES.extend(
        ['\tedit %s' % service,
         '\t\tset protocol TCP/UDP',
         '\t\tset %s-portrange %s' % (protocol.lower(), service),
         '\tnext']
        )
```

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+  def _generate_address_names(self, *addresses):
+    for group in addresses:
+      for addr in group:
+        if addr and not isinstance(addr, nacaddr.IPv6):
+          self._add_address_to_fw_addresses(addr.with_prefixlen)
+
+  def __str__(self):
+    lines = []
+
+    self._generate_address_names(self._term.destination_address, self._term.source_address)
```
ankenyr: Line too long.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+    # lines.extend(self.firewall_addresses)
+
+    dest_addresses = self._get_addresses_name(self._term.destination_address)
+    src_addresses = self._get_addresses_name(self._term.source_address)
+    services = self._get_services_string(self._term.protocol, self._term.destination_port)
```
ankenyr: line too long

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+
+    lines.append('\t\tset comments %s' % self._term.name)
+    lines.append('\t\tset srcintf %s' % (self._term.source_interface or 'any'))
+    lines.append('\t\tset dstintf %s' % (self._term.destination_interface or 'any'))
```
ankenyr: Line too long

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+  _PROTO_INT = True
+  _TERM_REMARK = True
+
+  def _BuildTokens(self):
+    """Build supported tokens for platform.
+
+    Returns:
+      tuple containing both supported tokens and sub tokens
+    """
+    supported_tokens, supported_sub_tokens = super(Fortigate, self)._BuildTokens()
```
ankenyr: Line too long

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+
+    # a mixed filter outputs both ipv4 and ipv6 acls in the same output file
+
+    for header, terms in pol.filters:
+      if self._PLATFORM not in header.platforms:
+        continue
+
+      filter_options = header.FilterOptions(self._PLATFORM)
+
+      if (len(filter_options) < 2 or filter_options[0] != "from-id"):
```
ankenyr: You switched to using double quotes here, please remain consistent. We use single quotes for all strings, see below where you use 'noverbose'

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+        raise UnsupportedFilterError(
+          "Fortigate Firewall filter arguments must specify from_id")
```
ankenyr: Same here

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+          raise FortiGateDuplicateTermError('You have a duplicate term: %s' %
+                                            term.name)
+        term_dup_check.add(term.name)
+
+        term.name = self.FixTermLength(term.name)
+        new_term = Term(term)
+
+        self.fortigate_policies.append((header, term.name, new_term))
+
+  def _GetTargetByPolicyID(self, id):
```
ankenyr: id is a built-in 
https://docs.python.org/3/library/functions.html#id

Please do not overwrite built-in methods as they can have unforeseen consequences.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+  def __str__(self):
+    start_addresses = ['config firewall address']
+    start_services = ['config firewall service custom']
+    start_policies = ['config firewall policy']
+    end = ['end']
+    target_addresses = []
+    target_services = []
+    target_policies = []
+
+    for (header, filter_name, term) in self.fortigate_policies:
```
ankenyr: header looks to be unused? If so please replace it with _ to mark it as unused.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+  _PROTO_MAP = {
+    'icmp': 'ALL_ICMP',
+    'gre': 'GRE',
+    'ip': 'ALL',
+    'tcp': _PORTS_TCP,
+    'udp': _PORTS_UDP
+  }
+
+  @staticmethod
+  def GetProtocol(protocol, port=None):
```
ankenyr: This method feels a bit messy. In a case where someone provides a l3 proto that doesn't match you would raise an error of "failed to get service from foo and none port" which doesn't really make sense to me.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+                         'logging'}
+
+    supported_sub_tokens.update({'option': {'from_id'},
+                                 # Warning, some of these are mapped
+                                 # differently. See _ACTION_TABLE
+                                 'action': {'accept', 'deny', 'reject',
+                                            'reject-with-tcp-rst'}})
+    return supported_tokens, supported_sub_tokens
+
+  def _TranslatePolicy(self, pol, exp_info):
```
ankenyr: Docstring here please.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+      if self._PLATFORM not in header.platforms:
+        continue
+
+      filter_options = header.FilterOptions(self._PLATFORM)
+
+      if (len(filter_options) < 2 or filter_options[0] != "from-id"):
+        raise UnsupportedFilterError(
+          "Fortigate Firewall filter arguments must specify from_id")
+
+      from_id = filter_options[1]
```
ankenyr: I have been rethinking options for a bit while I was on paternity. What exactly is the "from id" option for? Is it required for all policies? From the code it looks to be the header name but you add this into each term? and they increment?

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+      Term.CURRENT_ID = int(from_id)
+
+      self.verbose = True
```
ankenyr: Do you actually use verbose in here anywhere? This is used to toggle comments in the generated syntax.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+          if term.expiration <= current_date:
+            logging.warn('WARNING: Term %s in policy %s is expired and '
+                         'will not be rendered.', term.name, filter_name)
+            continue
+        if term.name in term_dup_check:
+          raise FortiGateDuplicateTermError('You have a duplicate term: %s' %
+                                            term.name)
+        term_dup_check.add(term.name)
+
+        term.name = self.FixTermLength(term.name)
```
ankenyr: Does fortigate actually have a length limit?

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+        new_term = Term(term)
+
+        self.fortigate_policies.append((header, term.name, new_term))
+
+  def _GetTargetByPolicyID(self, id):
+    return '\tedit %s' % id
+
+  def __str__(self):
+    start_addresses = ['config firewall address']
```
ankenyr: This section is a bit weird, is there a reason why you are not just appending things as they come to target?

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,330 @@
...
+
+    for (header, filter_name, term) in self.fortigate_policies:
+      target_policies.append(self._GetTargetByPolicyID(term.id))
+
+      term_str = str(term)
+
+      target_policies.append(term_str)
+
+      target_policies += ['\tnext', '']
+    target_addresses.extend(Term.get_fw_addresses())
```
ankenyr: I am unsure if this is a good idea to be using a static method like this for this purpose. You are already looping over the terms so collecting the addresses shouldn't be a problem. Storing a list of all the IP's in the class is also weird and feels incorrect to me.

capirca/lib/fortigate.py
Ali-aqrabawi: fortigate does not accept `port-numbers`, it only accept `service-name`.

capirca/lib/fortigate.py
Ali-aqrabawi: yes, in fortigate you should explicitly specify the ID/sequence for each policy, and this affect the firewall policies globally, so it should be specified by the user, as the user must have a preference where those policies should be inserted. so he specify the starting id and we can then start incrementing from there ,yeah Fortigate is really messy.

capirca/lib/fortigate.py
Ali-aqrabawi: it's `1023 char` so don't really need this, removing it

capirca/lib/fortigate.py
Ali-aqrabawi: those are lines needed to enter the config section of targets, for every target you need to add `addresses objects`, `service objects` and finally you can create the `policies`, we can append them directly to target, but this way we are grouping the addresses commands togather as well as services commands togather to have more readable config commands and less amount of commands.

capirca/lib/fortigate.py
Ali-aqrabawi: for same reason explained here:
 https://github.com/google/capirca/pull/173#discussion_r292202610

capirca/lib/fortigate.py
ankenyr: Are these sequences per header? as in can I have two sequences at 1 as long as they are for different directions/zones/whatever fortigate uses in its terminology? If so could they all start at 0 or 1 and increment based on their position? If these are not header dependent and a universal thing could they not be in the same sequence as they are in the file? Got any documentation I can read to get a clearer idea on this?

Ali-aqrabawi: no these sequence per `term/policy`, in each `term/policy` you have `source-interface` and `destination-interface` in addition to the other options, that's it you don't need to apply the policy you created anywhere else.

Ali-aqrabawi:
@ankenyr i have added pr and resolved the code reviews.
for https://github.com/google/capirca/pull/173#discussion_r292180314, i have create a `Container` object to hold the grouped commands like `service-objects` commands and `network-objects` commands

capirca/aclgen.py
```diff
@@ -52,6 +52,7 @@
 from capirca.lib import speedway
 from capirca.lib import srxlo
 from capirca.lib import windows_advfirewall
+from capirca.lib import fortigate
```
ankenyr: Please make sure imports re in order.

Ali-aqrabawi: noted

ankenyr:
Hi Ali, I have not heard anything on this PR in a while. There are still pending changes requested.

Ali-aqrabawi:
Hi Rob, i have done the requested changes [here](https://github.com/google/capirca/pull/173/commits/0662b9c49b3e5b29ac7573f2be4521fb39b18a89), did you check it?

ankenyr:
sorry but I don't see the comments as resolved which is why I never checked on anything.
https://imgur.com/KwyGS0O
Can you make sure all the comments are resolved and then I can make another pass?

Ali-aqrabawi:
done

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+
+"""Fortigate generator."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import datetime
+import six
```
ankenyr: Six should be after the capirca imports.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+
+from absl import logging
+from capirca.lib import aclgenerator
+from capirca.lib import nacaddr
+
+
+_ACTION_TABLE = {
+  'accept': 'accept',
```
ankenyr: add two spaces on these indents.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+
+class FortiGateFindServiceError(Error):
+  pass
+
+
+class FortiGateDuplicateTermError(Error):
+  pass
+
+
+class FortiGatePortDoesNotExist(Error):
```
ankenyr: Errors should end with the name error, see the above examples.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+    Converts a port number to a service name.
+    :param protocol: string representing protocol (tcp, udp, etc).
+    :param port: integer representing the port number.
+    :return: the service name of provided port-protocol
+    """
+    f_proto = FortigatePortMap._PROTO_MAP.get(protocol, None)
+    if f_proto is None:
+      raise FortiGateValueError(
+        '%r protocol is not supported by Fortigate, supported protocols = %r' % (
```
ankenyr: when continuing to a new line for parameters you should add two spaces to the indentation. Please fix this here as above and elsewhere in the code.

ankenyr: also this line is too long, 80 characters is the max on one line.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+        return f_proto[port]
+      except KeyError:
+        raise FortiGatePortDoesNotExist
+    else:
+      raise FortiGateFindServiceError(
+        'failed to get service from %r protocol and %r port' % (protocol, port)
+      )
+
+
+class ObjectsContainer:
```
ankenyr: did you mean
`class ObjectsContainer(object):`

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+  """a Container that holds service and network objects"""
+
+  def __init__(self):
+    self._FW_ADDRESSES = []
```
ankenyr: All capitals are usually reserved for constants. These are not that.
http://google.github.io/styleguide/pyguide.html#3164-guidelines-derived-from-guidos-recommendations

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+    """
+
+    services = []
+    if protocol and not ports:
+      services.append(FortigatePortMap.GetProtocol(protocol[0]))
+    for port in ports:
+      try:
+        service = FortigatePortMap.GetProtocol(protocol[0], port[0])
+      except FortiGatePortDoesNotExist:
+        self._obj_container._add_service_to_fw_services(protocol[0], port[0])
```
ankenyr: you are accessing a protected method of another object. This usually means that this method shouldn't be protected or that you shouldn't be doing that.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,383 @@
...
+
+  def __str__(self):
+    lines = []
+
+    self._generate_address_names(self._term.destination_address,
+                                 self._term.source_address)
+
+    dest_addresses = self._get_addresses_name(self._term.destination_address)
+    src_addresses = self._get_addresses_name(self._term.source_address)
+    all_ports = self.clean_ports(self._term.source_port, self._term.destination_port)
```
ankenyr: line too long, please fix this here and elsewhere.

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+
+"""Unittest for fortigate policy rendering module."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import string
+import re
```
ankenyr: re comes before string

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+      'protocol': 'tcp',
+      'src_addr': 'SOME_HOST',
+      'dest_addr': 'SOME_HOST',
+      'src_port': 'HTTP',
+      'dest_port': 'HTTP',
+      'action': 'accept',
+      'logging': 'true'
+  }
+
+  def format(*args, **kwargs):
```
ankenyr: This should have a bounded self argument right?

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+      return kwds[key]
+    except KeyError:
+      return self.DEFAULT_VALUES[key]
+
+
+EXP_INFO = 2
+
+
+class FortigateTest(unittest.TestCase):
+  def setUp(self):
```
ankenyr: one new line should be between new definitions.
http://google.github.io/styleguide/pyguide.html#35-blank-lines

also no call to super in the setUp, this should be done.

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+    self.naming = mock.create_autospec(naming.Naming)
+
+    def get_addr_side_eff(host):
+      hosts = {
+          'SOME_HOST': [nacaddr.IP('10.0.0.0/8')],
+          'SOME_HOST2': [nacaddr.IP('20.0.0.0/8')]
+      }
+      return hosts[host]
+
+    def get_port_side_eff(port, protocol):
```
ankenyr: protocol looks to be unused?

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+                      parsed_p,
+                      EXP_INFO)
+
+  def testAction(self):
+    accept_term = self.fmt.format(TERM_TEMPLATE, action='accept')
+    deny_term = self.fmt.format(TERM_TEMPLATE, action='deny')
+    reject_term = self.fmt.format(TERM_TEMPLATE, action='reject')
+
+    accept_acl = fortigate.Fortigate(
+      policy.ParsePolicy(GOOD_HEADER + accept_term,
```
ankenyr: Same indentation issue here

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+                         self.naming), EXP_INFO
+    )
+    reject_acl = fortigate.Fortigate(
+      policy.ParsePolicy(GOOD_HEADER + reject_term,
+                         self.naming), EXP_INFO
+    )
+
+    accept_sig = 'set action accept'
+    deny_sig = 'set action deny'
+    self.assertTrue(
```
ankenyr: There is a more specific assertIn function which could be used here I believe.

ankenyr: Fix this issue in other places where an seert In would be used.

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+      policy.ParsePolicy(GOOD_HEADER + custom_port_term,
+                         self.naming), EXP_INFO
+    )
+
+    dup_sig = 'set service HTTP\n'
+    diff_sig = 'set service HTTP HTTPS\n'
+    src_only_sig = dup_sig
+    icmp_sig = 'set service ALL_ICMP\n'
+    ip_sig = 'set service ALL\n'
+    custom_port_sig = 'config firewall service custom\n\tedit 43\n\t\t' \
```
ankenyr: do not use \ for line continuation.
http://google.github.io/styleguide/pyguide.html#32-line-length
The link provides ways to get around this.

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,390 @@
...
+      policy.ParsePolicy(GOOD_HEADER + no_log_term,
+                         self.naming), EXP_INFO
+    )
+
+    log_sig = 'set logtraffic all'
+
+    self.assertTrue(
+      log_sig in str(log_acl), '[%s]' % str(log_acl)
+    )
+    self.assertTrue(
```
ankenyr: AssertNotIn could be used here instead

ankenyr:
Hey @Ali-aqrabawi did you have time to fix the remaining issues here?

Ali-aqrabawi:
@ankenyr i'm working on it

Ali-aqrabawi:
@ankenyr everything should be fixed now

tests/lib/fortigate_test.py
Ali-aqrabawi: this is an overriding of `string.Formatter.format` so i did not modified the function parameters.

tests/lib/fortigate_test.py
Ali-aqrabawi: the super call is not needed in this case, [same here](https://github.com/Ali-aqrabawi/capirca/blob/cf91480d58e4a96d9f5e45f2609755d2c62a6c9c/tests/lib/cisco_test.py#L412) and for other `test modules`

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,394 @@
...
+"""Fortigate generator."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
+import datetime
+
+from capirca.lib import aclgenerator
```
ankenyr, line 25: absl logging belongs at the top here

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,394 @@
...
+
+    Args:
+      protocol: string representing protocol (tcp, udp, etc)
+      port: integer representing the port number
+
+    Returns:
+      string
+
+    Raises:
+      FortiGateValueError: When unsupported protocol is used.
```
ankenyr, line 136: Forgot FortigateServiceError in the docstring

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,394 @@
...
+  _PROTO_MAP = {
+      'icmp': 'ALL_ICMP',
+      'gre': 'GRE',
+      'ip': 'ALL',
+      'tcp': _PORTS_TCP,
+      'udp': _PORTS_UDP
+  }
+
+  @staticmethod
+  def GetProtocol(protocol, port=None):
```
ankenyr, line 125: GetProtocol should likely be get_protocol to conform with other method names in this file.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,394 @@
...
+      return f_proto
+    elif port:
+      try:
+        return f_proto[port]
+      except KeyError:
+        raise FortiGatePortDoesNotExistError
+    else:
+      raise FortiGateFindServiceError(
+          'service not found from %r protocol and %r port' % (protocol, port))
+
```
ankenyr, line 154: 2 blank lines between classes

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,394 @@
...
+class ObjectsContainer(object):
+  """A Container that holds service and network objects."""
+
+  def __init__(self):
+    self._fw_addresses = []
+    self._fw_services = []
+
+    self._FW_DUP_CHECK = set()
```
ankenyr, line 162: Not sure why this one is capitalized, it is not a constant.

capirca/lib/fortigate.py
```diff
@@ -0,0 +1,394 @@
...
+      all_ports += src_ports
+    if dest_ports:
+      all_ports += dest_ports
+    return set(all_ports)
+
+  def _get_services_string(self, protocols, ports):
+    """Get the service name, if not exist create it.
+
+    Args:
+      protocol: list of protocols
```
ankenyr, line 234: args here do not match the args above, need to add or move the "s" from above or below respectively

tests/lib/fortigate_test.py
ankenyr: https://docs.python.org/3.6/library/string.html#string.Formatter.format
Seems it takes a bound argument to me. What docs are you reading?

tests/lib/fortigate_test.py
ankenyr: New line is still required.

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,352 @@
...
+    self.assertIn(
+        dup_sig, str(dup_acl), '[%s]' % str(dup_acl))
+    self.assertIn(
+        diff_sig, str(diff_acl), '[%s]' % str(diff_acl))
+    self.assertIn(
+        src_only_sig, str(src_only_acl), '[%s]' % str(src_only_acl))
+    self.assertIn(
+        icmp_sig, str(icmp_acl), '[%s]' % str(icmp_acl))
+    self.assertIn(
+        ip_sig , str(ip_acl), '[%s]' % str(ip_acl))
```
ankenyr, line 264: extra space before the comma
ip_sig, str(ip_acl).....
        ^

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,352 @@
...
+                                                        'dest_interface'))
+    src_only_int_term = self.fmt.format(TERM_TEMPLATE,
+                                        src_interface='wan1',
+                                        remove_fields=('dest_interface',))
+    dest_only_int_term = self.fmt.format(TERM_TEMPLATE,
+                                         dest_interface='wan2',
+                                         remove_fields=('src_interface',))
+    both_interfaces_term = self.fmt.format(TERM_TEMPLATE,
+                                           src_interface='wan1',
+                                           dest_interface='wan2', )
```
ankenyr, line 280: Not sure why the end comma's here but at least on this line please remove the end space.

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,352 @@
...
+
+  def testPortMap(self):
+    port_map = fortigate.FortigatePortMap()
+    self.assertEqual('SSH', port_map.get_protocol('tcp', 22))
+    self.assertRaises(fortigate.FortiGatePortDoesNotExistError,
+                      port_map.get_protocol,
+                      'tcp', 5000)
+    self.assertRaises(fortigate.FortiGateValueError,
+                      port_map.get_protocol,
+                      'bad_proto', 22)
```
ankenyr, line 352: These tests are never actually run it seems. You forgot to include the if main section with a call to the unittests.

tests/lib/fortigate_test.py
ankenyr: Closed without resolution

tests/lib/fortigate_test.py
ankenyr: This was resolved without comment. In 3.6 self was defined within the method itself. In 3.8 it is in the method signature.
https://github.com/python/cpython/blob/3.8/Lib/string.py#L162

I also find this all kind of weird. You inherit from a class but then invoke the class method by its class identifier. Something like the following would be simpler here

```
  def Format(self, string, **kwargs):
    if 'remove_fields' in kwargs:
      for field in kwargs['remove_fields']:
        remove_regex = '.*' + field + '.*'
        string = re.sub(remove_regex, '', string)
    return self.format(string, **kwargs)
```

tests/lib/fortigate_test.py
ankenyr, line 352: Please add a test for build tokens to return as expected.

tests/lib/fortigate_test.py
```diff
@@ -0,0 +1,352 @@
...
+    accept_sig = 'set action accept'
+    deny_sig = 'set action deny'
+    self.assertIn(
+        accept_sig, str(accept_acl), '[%s]' % str(accept_acl))
+    self.assertIn(
+        deny_sig, str(deny_sig), '[%s]' % str(deny_acl))
+    self.assertIn(
+        deny_sig, str(reject_acl), '[%s]' % str(reject_acl))
+
+  def testAddresses(self):
```
ankenyr, line 166: You are collapsing a lot of checks into a single test with multiple unrelated asserts.
The output on a failure looks like the following

```
Traceback (most recent call last):
  File "tests/lib/fortigate_test.py", line 156, in testAction
    self.assertIn(
AssertionError: 'set action accept' not found in 'config firewall address\n\tedit 10.0.0.0/8\n\t\tset subnet 10.0.0.0/8\n\tnext\n \nend\n \nconfig firewall service custom\n \nend\n \nconfig firewall policy\n\tedit 2\n\t\tset comments good-term-2\n\t\tset srcintf wan1\n\t\tset dstintf wan2\n\t\tset dstaddr 10.0.0.0/8\n\t\tset srcaddr 10.0.0.0/8\n\t\tset action aaccept\n\t\tset service HTTP\n\t\tset schedule always\n\t\tset logtraffic all\n\tnext\n\nend' : [config firewall address
	edit 10.0.0.0/8
		set subnet 10.0.0.0/8
	next

end


end

config firewall service custom

end


end

config firewall policy
	edit 2
		set comments good-term-2
		set srcintf wan1
		set dstintf wan2
		set dstaddr 10.0.0.0/8
		set srcaddr 10.0.0.0/8
		set action aaccept
		set service HTTP
		set schedule always
		set logtraffic all
	next

end]

Ran 9 tests in 1.024s
```

Execution inside of a single test also stops at the first failure. If you had multiple issues such as both accept and deny failing for different reasons, this test would not report that both are wrong.

You may wish to look into paramaterization as this feels like what you are wanting to do. There has been talk about redoing a lot of our tests to use that.

capirca/lib/fortigate.py
ankenyr: docstrings should contain the arguments returns and other sections.
Take a look at some of the docstrings in junipersrx
https://github.com/google/capirca/blob/master/capirca/lib/junipersrx.py#L228
There are sections for the arguments, returns, methods, and attributes where applicable.

Please fix this here and elsewhere.

ankenyr:
Ali, there are still changes that have not been made that I have requested.

Ali-aqrabawi:
> Ali, there are still changes that have not been made that I have requested.

i checked the requested changes couple of times can't find which part i'm missing, sorry :/

Slepwin:
Any updates here?

nero85:
We are closing this in preference for #222. Thanks for your input!

Issue #180, opened 2019-12-12
Multiple verbatim lines are not processed in the cisco module


kaibabbob:
There is a return statement inside of the verbatim for loop (3 places) that prevents more than one verbatim line from being rendered into an ACL:

<snip>
    # Term verbatim output - this will skip over normal term creation
    # code by returning early.  Warnings provided in policy.py.
    if self.term.verbatim:
      for next_verbatim in self.term.verbatim:
        if next_verbatim[0] == self.platform:
          ret_str.append(' ' + str(next_verbatim[1]))
        return '\n'.join(ret_str)

    # protocol
</snip>

Here is a test policy:

header {
  comment:: "multiple verbatim line test"

  target:: cisco allowtointernet
}

term verbatim-test {
  comment:: "Testing verbatim"
  verbatim:: cisco "permit ip 10.0.0.0 255.0.0.0 any"
  verbatim:: cisco "permit ip 11.0.0.0 255.0.0.0 any"
}
Current rendering:

! $Id:$
! $Date:$
! $Revision:$
no ip access-list extended allowtointernet
ip access-list extended allowtointernet
 remark $Id:$
 remark multiple verbatim line test

 remark verbatim-test
 remark Testing verbatim
permit ip 10.0.0.0 255.0.0.0 any

exit

kaibabbob:
This has been resolved.

PR #183, opened 2020-03-13
Inside lib/cisco.py, a return statement was nested inside the for-loop instead of afterwards (in 3 places)

kaibabbob:
This fixes issue #180


kaibabbob:
@googlebot I signed it!

ankenyr:
Thanks for this, mind modifying the tests so it would catch this if it was reverted?

rdsharma:
If I'm reading correctly we already fixed this one upstream - if not, please reopen with a fresh rebase so that it can be picked up by our new tooling.

Issue #136, opened 2017-06-10
Terms that start with a DSCP match will cause parse errors.


robison:
Not 100% certain if this is intentional, but any terms that begin with a potential match for a DSCP marking followed by a hyphen will cause a parse error. e.g.:

    (venv) [13:03:56][ robisonjac-ltm:~/git/github.com/capirca ](✖ master) $ git diff tests/lib/cisco_test.py
    diff --git a/tests/lib/cisco_test.py b/tests/lib/cisco_test.py
    index 72d347b..cc4b918 100644
    --- a/tests/lib/cisco_test.py
    +++ b/tests/lib/cisco_test.py
    @@ -109,13 +109,13 @@ header {
     }
     """
     GOOD_STANDARD_TERM_1 = """
    -term standard-term-1 {
    +term cs5-standard-term-1 {
       address:: SOME_HOST
       action:: accept
     }
     """
     GOOD_STANDARD_TERM_2 = """
    -term standard-term-2 {
    +term cs5standard-term-2 {
       address:: SOME_HOST
       action:: accept
     }
    (venv) [13:04:01][ robisonjac-ltm:~/git/github.com/capirca ](✖ master) $

And upon running pytest, the following error is generated:

    p = LexToken(DSCP,u'cs5',7,84)

        def p_error(p):
          """."""
          next_token = yacc.token()
          if next_token is None:
            use_token = 'EOF'
          else:
            use_token = repr(next_token.value)

          if p:
            raise ParseError(' ERROR on "%s" (type %s, line %d, Next %s)'
    >                        % (p.value, p.type, p.lineno, use_token))
    E       ParseError:  ERROR on "cs5" (type DSCP, line 7, Next u'-')

    lib/policy.py:2328: ParseError

I can provide more details if needed, or would be happy to contribute a potential fix - but I wanted to be certain that this was unintended behavior.

ankenyr:
Hi @robison,
Sorry for letting this linger. I wouldn't say this it intended behavior. If you have a fix ready please send it and we can review it.

ankenyr:
Did you plan on making a fix for this @robison ?

PR #184, opened 2020-04-07
Fix-issue-136: any term beginning with a potential match for a DSCP marking followed by a hyphen will cause a parse error.

gmonni:
**PROBLEM STATEMENT**
CAPIRCA issue #136 : _any term beginning with a potential match for a DSCP marking followed by a hyphen will cause a parse error._
As an example both terms below cause  a match: 
- _cs4_ is a match (correct)
- _cs4-anyword_ is a match (**not** correct)

the parser interprets _cs4-anyword_ as a concatenation of two DSCP markers (_cs4_ and _anyword_) and then raises an Exception because '_anyword_' is an invalid dscp marker. 

**RCA**
if a term starts with what looks like a DSCP marker followed by a hyphen, the parser interprets the hyphen as a boundary between two DSCP markers.

**SOLUTION**
**When identifying a DSCP term we need to handle the '-' as part of the word, not as a boundary.** Detail explanation of the regex fix below:
 `Update the regex in t_DSCP: Add a negative lookahead for the hyphen to the regex in t_DSCP: this avoids matching a term that starts with a DSCP marker followed by a hyphen`. 
The new regex can be tested [here](https://regex101.com/r/kZfrFY/2/).
After this change,
- cs4 is a match (Correct)
- cs4-anyword is **not** a match (correct)

After applying the patch the parser will **not** interpret anymore _cs4-anyword_ as a concatenation of two DSCP markers.


gmonni:
@ankenyr this PR fixes #136 and adds the unit test. When you have the chance, could you please review it ?

Issue #187, opened 2020-05-30
Capirca doesn't allow use of 'is-fragment' under Cisco IOS


mpenning:
The [online docs](https://github.com/google/capirca/wiki/Policy-format) say that `is-fragment` is a valid option value for Cisco rules, but I get the following error when I use `is-fragments` for Cisco ACLs under Capirca=1.122:

```
Traceback (most recent call last):
  File "test_cisco_acl.py", line 37, in <module>
    cisco_acl = Cisco(pp, exp_info=0)
  File "/opt/virtual_env/py37_test/lib/python3.7/site-packages/capirca/lib/aclgenerator.py", line 325, in __init__
    raise UnsupportedFilterError('\n %s' % '\n'.join(all_err))
capirca.lib.aclgenerator.UnsupportedFilterError:
 test-deny-ip-fragments contains unsupported keywords (is-fragment) for target cisco in policy
```

Cisco IOS has options to catch IPv4 fragments, and this is a very important keyword for Cisco IOS ACLs (it's a known problem that you must deny IP fragments or people can potentially [circumvent your ACLs with fragments](https://www.cisco.com/c/en/us/support/docs/ip/access-lists/13608-21.html#anc34)).

Test script to reproduce...

```python
from capirca.lib.policy import ParsePolicy
from capirca.lib.naming import Naming

from capirca.lib.cisco import Cisco

defs = Naming()
defs.ParseServiceList([
    'SVC1 = 80/tcp 443/tcp',
    'SVC2 = 25/tcp',
    'ALL_SVC = SVC1 SVC2',
    ])
defs.ParseNetworkList([
    'LAN1 = 172.16.1.0/24',
    'LAN2 = 172.16.2.0/24',
    'LAN = LAN1 LAN2',
    ])

acl ="""
header {
  comment:: "Demo of capirca acl DSL"
  target:: cisco ACL_IN extended
}
term test-deny-ip-fragments {
  option:: is-fragment
  action:: deny
}
term allow-to-lan {
  destination-address:: LAN
  protocol:: tcp
  source-port:: ALL_SVC
  action:: accept
}
"""

pp = ParsePolicy(acl, defs)
cisco_acl = Cisco(pp, exp_info=0)
print(cisco_acl)
```

PR #188, opened 2020-05-30
Fix Github Issue #187 with max backwards-compat (retaining support fo…

mpenning:
…r 'fragments')

I added a new `is-fragment` Cisco IOS ACL term option while retaining the old `fragments` option.  AFAICT, `is-fragment` is the preferred syntax (at least that's what the docs say).

### DIFF
--- a/capirca/lib/cisco.py
+++ b/capirca/lib/cisco.py
@@ -735,7 +735,8 @@ class Term(aclgenerator.Term):
         and ('tcp-established' in opts or 'established' in opts)):
       if 'established' not in self.options:
         self.options.append('established')
-    if ('ip' in protocol) and ('fragments' in opts):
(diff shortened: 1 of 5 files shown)


capirca/lib/cisco.py
```diff
@@ -735,7 +735,8 @@ def __str__(self):
         and ('tcp-established' in opts or 'established' in opts)):
       if 'established' not in self.options:
         self.options.append('established')
-    if ('ip' in protocol) and ('fragments' in opts):
+    # Using both 'fragments' and 'is-fragment', ref Github Issue #187
+    if ('ip' in protocol) and (('fragments' in opts) or ('is-fragment' in opts)):
```
ankenyr, line 739: This line would be too long, there is an 80 character line limit.

PR #190, opened 2020-06-02
Condensed policy rendering logic

taliamax:
The function `RenderFile` uses a lot of repeated logic. This PR condenses that logic by using a dictionary to contain all the data instead of individual variables so that it can be looped through instead of manually parsed.

### DIFF
--- a/capirca/lib/juniper.py
+++ b/capirca/lib/juniper.py
@@ -269,9 +269,14 @@ class Term(aclgenerator.Term):
         else:
           from_str.append('%s;' % opt)
 
-    # term name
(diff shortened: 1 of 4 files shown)


ankenyr:
Thanks for this contribution, I was actually trying to think of ways to get rid of that section and somehow dynamically load the correct libraries. The issue is internally we have some libraries that are not exported. I am sure others have their own libraries that they do not want to export either.

Is there a way we could fix things so that anything that is a platform is automatically loaded and the deep copies are performed for them without explicit writing of each platform as we have it now?

PR #192, opened 2020-06-05
Fixed deprecation warnings

taliamax:
When running unit tests, there are many deprecation warnings due to legacy code from python2. This eliminates those warnings so that error messages from tests are more meaningful.


taliamax:
Force pushed after rebase to fix the merge conflicts

PR #193, opened 2020-06-06
Fixed tests and run script

taliamax:
- Fixed typo on variable name which made the aclgen script unrunnable
- Added import for missing module in the aclgen test
- Fixed issue with absl flags on aclgen test
- Upped python version for travis test
- Added minimum package version for 'six' module which was outdated on travis and causing tests to fail
- Added previously removed `entry_point` function to fix broken install (#194)
- Added test for `entry_point` function to ensure that tests fail if it is removed in the future


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: python
 python:
-  - "2.7"
+  - "3.7"
```
rdsharma, line 3: We run on 3.6 internally right now and there are some language changes in 3.7 like extra reserved keywords and breaking stdlib changes. Maybe we should test on both 3.6 and 3.7 for compatibility reasons?

tests/integration/aclgen_test.py
```diff
@@ -31,7 +31,7 @@
 
 
 FLAGS = flags.FLAGS
-FLAGS(sys.argv)
+FLAGS(sys.argv[0:1])
```
rdsharma, line 34: Can we comment this?

capirca/aclgen.py
```diff
@@ -55,53 +55,56 @@
 
 
 FLAGS = flags.FLAGS
-flags.DEFINE_string(
+
+
+def SetupFlags():
```
taliamax, line 60: External applications that also use `absl` can hit name conflicts when importing this file. Containing the setup in it's own function allows us to do the setup only when necessary, instead of at import time

tests/integration/aclgen_test.py
taliamax: Done!

requirements.txt
```diff
@@ -6,4 +6,4 @@ ipaddress>=1.0.22
 absl-py
 ply
 mock
-six # requirement from absl-py
+six>=1.12.0
```
taliamax, line 9: I think it'd be good to use frozen package versions in the requirements file to avoid future similar issues and help with duplicating environments, but that's outside of the scope of this PR

tests/integration/aclgen_test.py
```diff
@@ -43,6 +50,7 @@ def setUp(self):
     shutil.copytree('policies', self.pol_dir)
     self.context = multiprocessing.get_context()
 
+
```
taliamax, line 53: 2 lines between functions is compliant with PEP8

tests/integration/aclgen_test.py
```diff
@@ -59,6 +67,7 @@ def test_smoke_test_generates_successfully(self, mock_writer):
         os.path.join(self.test_subdirectory, f), mock.ANY) for f in files]
     mock_writer.assert_has_calls(expected, any_order=True)
 
+
```
taliamax, line 70: 2 lines between functions is compliant with PEP8

taliamax:
This was merged through https://github.com/google/capirca/commit/89a9c15153ab880833fe953a4ecd3455d2c07f84 - should this be closed now?

PR #201, opened 2020-06-18
Added single-threaded logic

taliamax:
When `max_renderers` is set to 1, only a single thread will be running anyways and we can forego the overhead of the multiprocessing pool. Plus, cProfile doesn't play nicely with multiprocessed applications and fails due to a pickling error.

### DIFF
--- a/capirca/aclgen.py
+++ b/capirca/aclgen.py
@@ -106,8 +106,20 @@ def SetupFlags():
       'Raise an error when a term is completely shaded by a prior term.')
   flags.DEFINE_integer(
       'exp_info',
-      2,
+       2,
       'Print a info message when a term is set to expire in that many weeks.')
+  flags.DEFINE_boolean(
+      'profile',
+      False,
+      'Run a thread to profile the execution. Implies \'--max_renderers 1\'')
+  flags.DEFINE_integer(
+      'profile_time',
+      None,
+      'The duration (seconds) that the profile thread should run for. Implies --profile.\n(default: 30)\n(an integer)')
+  flags.DEFINE_string(
+      'pprof_file',
+      None,
+      'The name of the output file for the profile thread. Implies --profile.\n(default:  \'profile.pprof\')')
 
 
 class Error(Exception):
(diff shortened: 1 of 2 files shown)


capirca/aclgen.py
```diff
@@ -463,6 +463,19 @@ def _WriteFile(output_file, file_string):
     raise
 
 
+def DiscoverAllPolicies(base_directory, output_directory, definitions):
+  logging.info('finding policies...')
+  pols = []
+  pols.extend(
```
rdsharma, line 491: Nitty nit: I think it's slightly more pythonic to write pols = [].extend(... ? But maybe I"m wrong?

capirca/aclgen.py
```diff
@@ -482,23 +495,46 @@ def Run(base_directory, definitions_directory, policy_file, output_directory,
...
     RenderFile(base_directory, policy_file, output_directory, definitions,
                FLAGS.exp_info, write_files)
+  elif FLAGS.max_renderers == 1:
+    # If only one process, run it sequentially
+    policies = DiscoverAllPolicies(
+      base_directory,
+      output_directory,
+      definitions
+    )
+    result = []
```
rdsharma, line 527: Is this variable used?

capirca/aclgen.py
```diff
@@ -482,23 +495,46 @@ def Run(base_directory, definitions_directory, policy_file, output_directory,
...
-    pols.extend(DescendRecursively(base_directory, output_directory,
-                                   definitions))
-
+    policies = DiscoverAllPolicies(
+      base_directory,
+      output_directory,
+      definitions
+    )
     pool = context.Pool(processes=FLAGS.max_renderers)
     results = []
```
rdsharma, line 545: Also seems unused (in original code)

taliamax: It's used to store all the promises, I think, but I'm not 100% sure why

capirca/aclgen.py
taliamax: `extend` is a mutating function so it returns None, which means we can't use it in that way

PR #203, opened 2020-06-29
new: addded iputils for faster address exclusion

taliamax:
The default python standard library `address_exclude` function uses an O(n) algorithm for excluding addresses, and returns the list of addresses in reverse sorted order. This algorithm performs the exclusion in O(1) and returns the addresses in sorted order.


capirca/utils/iputils.py
```diff
@@ -0,0 +1,52 @@
...
+                 that is being removed from the base_net
+  raises:
+    ValueError if exclude_net is not completely contained in base_net
+  '''
+
+  if not base_net._version == exclude_net._version:
+    raise TypeError("%s and %s are not of the same version" % (
+      base_net, exclude_net))
+
+  if not isinstance(exclude_net, _BaseNetwork):
```
rdsharma: Should we check this first, before we access any members? Do we need to check the type of base_net as well?

taliamax: True! With the stdlib code it's not necessary since it's being called from the class but with this one that's a good extra check

PR #208, opened 2020-07-20
new: more strict naming for service definitions

taliamax:
Added more strict rules to service definition naming, and used these rules to infer if a token is an IP or a service during GetNet

### DIFF
--- a/capirca/lib/naming.py
+++ b/capirca/lib/naming.py
@@ -53,6 +53,8 @@ import glob
 import os
 import re
 
+from absl import logging
+
 from capirca.lib import nacaddr
 from capirca.lib import port as portlib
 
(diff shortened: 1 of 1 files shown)


rdsharma:
Did we decide to abandon this one?

taliamax:
We had a small conversation on whether it should still be around. I think it would be advisable to keep it but remove the strict requirement, since if we allow it as a first-pass we will still add some efficiency

capirca/lib/naming.py
```diff
@@ -134,6 +134,7 @@ def __init__(self, naming_dir=None, naming_file=None, naming_type=None):
     self.unseen_networks = {}
     self.port_re = re.compile(r'(^\d+-\d+|^\d+)\/\w+$|^[\w\d-]+$',
                               re.IGNORECASE|re.DOTALL)
+    self.token_re = re.compile(r'(^[-_A-Z0-9]+$)')
```
rdsharma, line 137: Do we disallow uppercase on purpose? (I don't see a re.IGNORECASE here, or we could just do A-z instead of A-Z)

rdsharma: Friendly ping :)

taliamax: Whoops! I thought I replied to this. I'm updating this PR today

PR #209, opened 2020-07-20
fix: fixed indentation on tests

taliamax:
These two tests were indented improperly and that was causing some issues


taliamax:
Whoops turns out my local just hadn't updated for some reason

PR #211, opened 2020-08-07
Feature/auto loaded modules

taliamax:
- Restructured Capirca files so they no longer exist in just a flat `lib/` folder. This is a breaking change for any system built on top of Capirca (as now different packages need to be used), but will inevitably assist in maintaining this moving forward as it will be clear where files should live.
- Added auto-loading of Capirca platforms to the `aclgen` command, which means that any platforms that are supported through that command can be added directly to the `capirca.platforms.exported` package, and provided that they export the two common variables used, they will be auto-detected when running `aclgen`


taliamax:
Rebased!

taliamax:
i think it might be better to just abandon this PR....

PR #212, opened 2020-08-11
fix: removed google-specific line causing errors

taliamax:
This should fix the issue that the CI is having right now with capirca


rdsharma:
We don't need this one anymore, closing

PR #213, opened 2020-08-13
new: added file-base configuration for capirca

taliamax:
Allow the usage of YAML files to configure `aclgen` for Capirca.

Despite the default being stripped from the CLI definition and inserted into the description, there should be no difference between the previous `aclgen --helpfull` to the one that implements this pull request. This is because all config defaults have now been moved to `capirca.utils.config`. Changes to the defaults in that module will translate to changes in the default flags.

### DIFF
--- a/capirca/aclgen.py
+++ b/capirca/aclgen.py
@@ -116,8 +116,13 @@ def SetupFlags():
     % str(config.defaults['shade_check']).lower())
   flags.DEFINE_integer(
     'exp_info',
-    2,
-    'Print a info message when a term is set to expire in that many weeks.')
+    None,
+    'Print a info message when a term is set to expire in that many weeks.\n(default: \'%s\')'
+    % str(config.defaults['exp_info']))
+  flags.DEFINE_multi_string(
+    'config_file',
+    None,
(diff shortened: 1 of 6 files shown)


capirca/aclgen.py
```diff
@@ -53,61 +53,76 @@
...
 
 
 def SetupFlags():
   flags.DEFINE_string(
-      'base_directory',
-      './policies',
-      'The base directory to look for acls; '
-      'typically where you\'d find ./corp and ./prod')
+    'base_directory',
+    None,
```
maddychan, line 65: It seems you've changed all the defaults to None; would this potentially cause problems with people omitting a flag because they wanted the default value?

taliamax: The functionality is maintained through the configs.default object. Down below in the main function, there is a line that extracts just the set values and applies them over the defaults, and the defaults are preserved as the same as before

Zenix27 (approved):

taliamax:
So it seems I forgot to remove the config flags from the `feature/zprofiling` branch before merging it in, but since the actual profiling logic was scrapped those flags don't really do anything 😅 removed them in this PR and rebased

PR #214, opened 2020-08-19
chg: refactored aclgen for legibility

taliamax:
* Used pathlib instead of manipulating path strings with the os module (compatibility: python3.4+)
* Removed SkipLines function and implemented the same functionality through the more concise `filter` function
* Changed named lambdas into function declarations (pep8 E731)


capirca/aclgen.py
```diff
@@ -387,70 +388,61 @@ def FilesUpdated(file_name, new_text, binary):
...
+  for directory in policy_directories:
+    directory_policies = list(directory.glob('*.pol'))
+    depth = len(directory.parents) - 1
+    logging.warning(
+      '-' * (2 * depth) + '> %s (%d pol files found)'
+      % (directory, len(directory_policies))
+    )
+    policy_files.extend(filter(lambda path: path.is_file(), directory_policies))
+
+  return policy_files
```
taliamax, line 444: this could be done more succinctly by globbing for the `.pol` files instead of `/pol` directories, but that would make the logging output harder to achieve

PR #220, opened 2020-10-13
Fixes #217

### LINKED ISSUE #217: Make the use of repository tags configurable/optional
After the discussion in Slack I'm proposing to make the use of the (only? with perforce/cvs useful) repository tags configurable via a parameter in the `target::` line of a policy.
I guess a similar approach as the `not-interface-specific` in juniper.py should do the trick. To not break backwards compatibility I would assume the current behavior as default and drop the repository tags only if the parameter is present. 

What do you think?

fischa:
This introduces a new parameter for the header section of juniper to drop the p4 repository tags if present. Default behaviour is to show them to not break current behaviour. 

This was tested with `target:: juniper LOOPBACK no-repository-tags` in the sample_juniper_loopback.pol file and worked as expected.

When running flake8 as recommended it shows some errors/warnings unrelated to the changes I made. One related error didn't make sense to me but I'm happy to fix if cleared up.

Feedback welcome. Let me know if I'm missing something.

### DIFF
--- a/tests/lib/arista_tp_test.py
+++ b/tests/lib/arista_tp_test.py
@@ -1061,6 +1061,9 @@ class AristaTpTest(unittest.TestCase):
         self.assertIn("match MIXED_INET ipv4", output, output)
         self.assertIn("source prefix 8.8.4.4/32", output, output)
         self.assertIn("destination prefix 10.0.0.0/8", output, output)
+        self.assertNotIn("match ipv6-MIXED_INET ipv6", output, output)
+        self.assertNotIn("source prefix 2001:4860:4860::8844/128",
+                         output, output)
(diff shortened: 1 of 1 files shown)


fischa:
I've split the repository tags as requested and used them in the `AddRepositoryTags()` as parameters.

fischa:
Can someone please take a look and comment? I'm happy to change things if needed.

greenpau:
@fischa , please rebase your PR.

fischa:
@greenpau I tried a rebase but I'm not sure if the result is what you wanted/expected. Also I see conflicting files I didn't touch with my PR. A pointer/hint would be appreciated.

PR #231, opened 2021-01-08
add arista traffic-policy support

sulrich:
this PR provides initial support for arista traffic-policies.  traffic-policies are a relatively new ACL/filter mechanism supported on recent arista platforms. 

support is provided for the following capirca tokens: 
 - `action`
 - `comment`
 - `counter`
 - `destination-address`
 - `destination-exclude`
 - `destination-port`
 - `destination-prefix` 
 - `fragment-offset`
 - `icmp-type`
 - `logging`
- `option`
   - `established`
   - `tcp-established`
   - `initial`
   - `rst`
   - `first-fragment`
 - `packet-length`
 - `source-address`
 - `source-exclude`
 - `source-port`
 - `source-prefix`
 - `verbatim`

tests, some initial documentation regarding rendering, and sample capirca policy files (one sample specific to traffic-policies and the addition of traffic-policies to the sample_multitarget.pol) are also included in this PR.

this addition is **not** an arista product


sulrich:
@googlebot I signed it!

sulrich:
hello!  just wondering if there's someone i could potentially work with on this particular PR?

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+            self.lines.append(line)
+            return
+
+        self.lines.append(line_indent + line.strip())
+
+
+class Term(aclgenerator.Term):
+    """represents an individual AristaTrafficPolicy term.
+
+    useful for the __str__() method. wher literally, everything intersting
```
abhindes, line 116: nit: where, interesting

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+            "established": "established",
+            # tcp flags
+            "ack": "ack",
+            "fin": "fin",
+            "initial": "initial",
+            "psh": "syn",
+            "rst": "rst",
+            "syn": "syn",
+            "tcp-est": "established",
+            "urg": "urg",
```
abhindes, line 163: All the TCP flags seem to be the same between inet and inet6, is there a reason you added them here?

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+        },
+        "inet6": {
+            "addr_fam": "ipv6",
+            "addr": "address",
+            "saddr": "source prefix",
+            "daddr": "destination prefix",
+            "protocol": "protocol",
+            # tcp flags
```
abhindes, line 171: missing "established" here.

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+    # encountered, they should be added to this table.  Accessing members
+    # of this table looks like:
+    #  self._TERM_TYPE('inet').get('saddr') -> 'source-address'
+    #
+    # it's critical that the members of each filter type be the same, that is
+    # to say that if _TERM_TYPE.get('inet').get('foo') returns something,
+    # _TERM_TYPE.get('inet6').get('foo') must return the inet6 equivalent.
+    _TERM_TYPE = {
+        "inet": {
+            "addr_fam": "ipv4",
```
abhindes, line 149: none of the other terms, other than "addr_fam" probably need to be here, since they all seem to be the same, unless there is a reason to expect that they will change, in which case, please ignore this comment.

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+            if self.term.protocol_except:
+                protocol_str = self._processProtocolExcept(self.term_type,
+                                                           self.term, flags)
+
+            # tcp/udp port generation
+            port_str = self._processPorts(self.term)
+            if port_str != "":
+                protocol_str += port_str
+
+            # icmp[v6] handling
```
abhindes, line 380: You'll want to add a check here to ensure that icmp is only rendered when the term contains actual IPv4 addresses only, and icmpv6 is rendered when there's only IPv6 addresses.

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+                    "ttl %s" % self.term.ttl,
+                    False
+                ])
+
+            if len(misc_options) > 0:
+                for mopt in misc_options:
+                    term_block.append([MATCH_INDENT, mopt, False])
+
+        # ACTION HANDLING
+        # if there's no action, then this is an implicit permit
```
abhindes, line 430: Could you add this bit of information in the action section of arista_tp.md.

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+                "udp": "",        # UDP
+                "vrrp": "",       # VRRP (112)
+            }
+        }
+
+        protocol_str = ""
+        prots = []
+        # if there are dirty prots we'll need to convert the protocol list to
+        # all numbers and generate the list of protocols to match on. EOS
+        # doesn't support commingling named protocols w/numeric protocol-ids
```
abhindes, line 540: What does the device return if you do a "show config" type of command? 
When I input a:
1. protocol name
2. protocol numeric-id for a protocol that has a well known protocol name
3. protocol numeric-id for a protocol that doesn't have a name.

In order to meet this requirement [0], I'd recommend using what the device expects.

[0] - https://github.com/google/capirca/blob/master/doc/generator_patterns.md#syntax-of-the-config-from-the-generator-and-on-device-should-match-when-cryptographically-verified

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+
+        if self.term.icmp_code and len(icmp_types) <= 1:
+            icmp_codes = self._Group(self.term.icmp_code)
+            icmp_codes = re.sub(r" ", ",", icmp_codes)
+            icmp_code_str += " code %s" % icmp_codes
+
+        return icmp_type_str, icmp_code_str
+
+    def _processProtocol(self, term_type, term, flags):
+        _ANET_PROTO_MAP = {
```
abhindes, line 510: Is this list exhaustive for the named protocols for arista traffic-policies?

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+        )
+
+        term_af = self.AF_MAP.get(self.term_type)
+        # comment - could use a little formatting love
+        if self.term.owner and not self.noverbose:
+            self.term.comment.append("owner: %s" % self.term.owner)
+        if self.term.comment and not self.noverbose:
+            for comment in self.term.comment:
+                for line in comment.split("\n"):
+                    term_block.append([MATCH_INDENT, "!! " + line, False])
```
abhindes, line 267: Are these style of comments (!!) retained in the policy if I do a "show config" style command?

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+             ]
+        )
+
+        term_af = self.AF_MAP.get(self.term_type)
+        # comment - could use a little formatting love
+        if self.term.owner and not self.noverbose:
+            self.term.comment.append("owner: %s" % self.term.owner)
+        if self.term.comment and not self.noverbose:
+            for comment in self.term.comment:
+                for line in comment.split("\n"):
```
abhindes, line 266: This could do with some sort of comment length truncation/splitting into multi-line comments. 
Large comments would end up in a single indented line.

Is there a comment width limit? If so, you can split on that basis, otherwise, do what is native to the platform for multi-line comments. If there's nothing specified by the platform, I would recommend splitting after 80 characters.

abhindes: https://github.com/google/capirca/blob/master/doc/generator_patterns.md#perform-truncating-of-names-for-term-name-or-comment-based-on-max-length is the corresponding requirement.

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+
+            for pol_term in terms:
+                for ft in ftypes:
+                    if filter_type == "mixed":
+                        term = copy.deepcopy(pol_term)
+                    else:
+                        term = pol_term
+
+                    term_type = ft
+                    term.name = self.FixTermLength(term.name)
```
abhindes, line 838: Is there a limit to the term-name length specified by the platform? 

Currently this is set to the default of 62, after which FixTermLength throws an error.

abhindes: If you end up having to truncate term names, that could result in an issue as well. 

Corresponding requirement - https://github.com/google/capirca/blob/master/doc/generator_patterns.md#perform-truncating-of-names-for-term-name-or-comment-based-on-max-length

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+            "name",
+            "option",
+            "owner",
+            "packet_length",
+            "platform",
+            "platform_exclude",
+            "port",
+            "protocol",
+            "protocol_except",
+            "source_address",
```
abhindes, line 727: please add "source_address_exclude" here.

doc/arista_tp.md
```diff
@@ -0,0 +1,55 @@
...
+The fully supported actions are: `accept`, and `deny`.  Use of `reject`, or `reject-with-tcp-rst` will result in the generation of deny actions in the rendered traffic policy.
+
+### counters
+- If counters are specified in a term, a traffic-policy named-counter stanza will be generated in the rendered output.
+- Counter names should not contain a (`.`). If a (`.`) is embedded in a counter name it will be replaced w/a dash (`-`).
+
+### (source|destination)-address-exclude
+Currently, (as of Jan-2021), EOS does not support the use of 'except' inline within match statements.  If an exclude/except token is used, a traffic-policy field-set will be generated and referenced in the match-term output. This field-set will be named `<direction>-<term.name>` where direction is either **src** or **dst** depending on the direction of the token in use.
+
+If the filter type is mixed, both address-families will have the respective field-sets generated. The field-set for the ipv4 address family will have the field-set generated with no suffix, while the ipv6 field-set will have `_v6` appended.
```
abhindes, line 44: nit: The ipv6 would have been more consistent if it had the 'ipv6-' added as a prefix, rather than a suffix. Similar to the the ipv6-default-rule. But this is a nitty nit, please feel free to ignore.

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+        # if accept and there are extra actions generate an actions statement
+        # if accept and no extra actions don't generate an actions statement
+        if self.term.action != ["accept"]:
+            term_block.append([MATCH_INDENT, "actions", False])
+            term_block.append([ACTION_INDENT, "%s" % current_action, False])
+        elif self.term.action == ["accept"] and has_extra_actions:
+            term_block.append([MATCH_INDENT, "actions", False])
+
+        if has_extra_actions:
+            # logging
```
abhindes, line 449: Could you do the same for 'syslog' as well, according to the traffic-policy config the log gets logged in the syslog :) 

Actually, to meet this requirement [0], it'd be best to add all positive logging actions here, and enable logging for them. So for _LOGGING = set(('true', 'True', 'syslog', 'local', 'log-both')), set logging. Only for 'disable', do nothing. 

[0] - https://github.com/google/capirca/blob/master/doc/generator_patterns.md#logging-is-supported-correctly-for-different-types-of-logging

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+                                " ".join(unsupported_opts),
+                            )
+                            continue
+
+                    has_unsupported_match_criteria = (
+                        term.dscp_except or
+                        term.dscp_match or
+                        term.ether_type or
+                        term.flexible_match_range or
+                        term.forwarding_class or
```
abhindes, line 896: Would forward_class be what could implement traffic-policy's traffic class? Not sure if this is super urgent, or a must-have kinda feature, but I saw that traffic classes are supported. I believe that means the same as forwarding classes, except in Juniper, they're BE, AF etc, and in Arista they are 0-7.

doc/arista_tp.md
```diff
@@ -0,0 +1,55 @@
...
+- `option`
+   - `established`
+   - `tcp-established`
+   - `initial`
+   - `rst`
+   - `first-fragment` - this  will be rendered as a `fragment` match.`
+ - `packet-length`
+ - `source-address`
+ - `source-exclude`
+ - `source-port`
```
abhindes, line 25: Can you add a section here that says # ports, and add the text -

In EOS traffic-policies, ports can be configured using:
source port <port-list> | field-set <port-set-name> ]
destination port <port-list> | field-set <port-set-name> ]
Currently, field-sets are not supported for ports, and only port lists are supported.

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+# generic error class
+class Error(Exception):
+    pass
+
+
+class TcpEstablishedWithNonTcpError(Error):
+    pass
+
+
+class AristaTpDuplicateTermError(Error):
```
rdsharma, line 70: Please reuse existing error classes wherever possible. We already have a DuplicateTermError in aclgenerator.py:
https://github.com/google/capirca/blob/188fd110dbf4358f4f46a730d18f49ccc1981e53/capirca/lib/aclgenerator.py#L58

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+                try:
+                    num_prots.append(str(self.PROTO_MAP[p]))
+                except KeyError:
+                    num_prots.append(str(p))
+            protocol_str += "protocol %s" % ",".join(num_prots)
+        else:
+            protocol_str += "protocol %s" % self._Group(prots)
+
+        if prots == ["tcp"]:
+            if len(flags) > 0:
```
rdsharma, line 561: Nit: combine if statements with "and"

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+        for p in except_list:
+            if 255 > p > ptr:
+                if (p - 1) == ptr:
+                    ex_str += str(ptr) + ","
+                    ptr = p + 1
+                else:
+                    ex_str += str(ptr) + "-" + str(p - 1) + ","
+                    ptr = p + 1
+            elif p == ptr:
+                ptr = p + 1
```
rdsharma, line 591: Can you hoist out `ptr = p + 1 `

capirca/lib/arista_tp.py
```diff
@@ -0,0 +1,1036 @@
...
+
+        for opt in [str(x) for x in options]:
+            # note: traffic policies support additional tcp flags. for now,
+            # only handle the required elements
+            #
+            # only append tcp-established for option established when
+            # tcp is the only protocol
+            if opt.startswith("established"):
+                if self.term.protocol == ["tcp"]:
+                    if "established" not in flags:
```
rdsharma, line 610: Combine if statements

rdsharma (changes requested):
Overall looks good, thanks for writing all those tests!

rdsharma:
Please rebase to resolve conflicts

sulrich:
ack - rebased. resolved conflict with `doc/policy_format.txt`

rdsharma (approved):

abhindes (approved):
All my comments have been addressed.

PR #239, opened 2021-01-19
PA: Add field pan-security-profile-group to policy.

hkam40k:
Add support for using "Security Profiles" on terms on Palo Alto platform.


greenpau:
@hkam40k , currently working on re-writing the generator. I hope will have something next week. This will need to way until then.

PR #244, opened 2021-03-09
Convert hand rolled XML document generation to use the ElementTree module.

kevinsteves:
Functionally equivalent to previous with exceptions:
  - remove XML declaration as not needed.
  - change config version to 8.1.0 (earliest supported PAN-OS)

Probable bug and question in XXX comments.  Can address these in
separate commits as needed.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -16,6 +16,7 @@
 import collections
 import datetime
 import logging
+import xml.etree.ElementTree as ET
 
 from capirca.lib import aclgenerator
 from capirca.lib import nacaddr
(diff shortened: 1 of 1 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -466,170 +467,148 @@ def _BuildPort(self, ports):
...
 
-        rules.append(self.INDENT * 9 + "<action>" + Term.ACTIONS.get(
-            str(options["action"])) + "</action>")
+        action = ET.SubElement(entry, "action")
+        action.text = Term.ACTIONS.get(options["action"])
 
+        # XXX loop vars, > 1 zones case
         if fz == tz == "any":
-          rules.append(self.INDENT * 9 + "<rule-type>interzone</rule-type>")
+          rule_type = ET.SubElement(entry, "rule-type")
```
greenpau, line 549: @kevinsteves , is `rule-type` it a valid field for 10+?

rdsharma:
Can we convert the tests verify the element tree itself rather than the rendered XML?

capirca/lib/paloaltofw.py
```diff
@@ -466,170 +467,148 @@ def _BuildPort(self, ports):
...
+          for x in options["service"]:
+            member = ET.SubElement(service, "member")
+            member.text = x
 
-        rules.append(self.INDENT * 9 + "<action>" + Term.ACTIONS.get(
-            str(options["action"])) + "</action>")
+        action = ET.SubElement(entry, "action")
+        action.text = Term.ACTIONS.get(options["action"])
 
+        # XXX loop vars, > 1 zones case
```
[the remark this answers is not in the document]
kevinsteves: XXX is used to identify an area to revisit, something that needs
attention, something you're unsure about, etc.

The commit message comments on the 2 XXX uses.

capirca/lib/paloaltofw.py
kevinsteves: yes, it was added in 6.1 I think.

admin@TEST-HOST5# find command keyword rule-type
set rulebase security rules <name> rule-type <universal|intrazone|interzone>
[edit] 
admin@TEST-HOST5# run show system info | match sw-version
sw-version: 10.0.4
[edit]

kevinsteves:
> Can we convert the tests verify the element tree itself rather than the rendered XML?

Eventually I think it does make sense to make the ElementTree root
available for applications like the tests.  Today there are 4 tests
that match XML and 2 or 3 likely change with ICMP enhancements/fixes.

greenpau:
@kevinsteves , please sync with the latest changes.

kevinsteves:
close.  will submit new PR on new target branch.

PR #251, opened 2021-03-17
Additional tcp-flag filters for juniper

britchey80:
Currently all of these tcp-flag filters (fin, psh and urg) are unusable with capirca.


britchey80:
@googlebot I signed it!

greenpau:
@britchey80 , what does your “git log” show in terms of email and name?

pims848:
Hello, We also would like to add more tcp-flags support in Juniper but we need some flag combinations ie: `tcp-flags "!(syn&ack)"`
Shall we hard code them or build something more powerful?

PR #253, opened 2021-03-18
Convert hand rolled XML document generation to use the ElementTree module.

kevinsteves:
Functionally equivalent to previous with exceptions:

    remove XML declaration as not needed

    change config version to 8.1.0 (earliest supported PAN-OS)

    don't indent XML as this can be performed with other tooling
    (e.g., xmllint)

    application groups are not currently used and the XML was
    not rendered correctly, so remove XML generation for now.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -18,6 +18,7 @@ import collections
 import datetime
 import logging
 import re
+from xml.dom import minidom
 import xml.etree.ElementTree as etree
 from capirca.lib import aclgenerator
 from capirca.lib import nacaddr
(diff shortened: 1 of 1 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -726,209 +726,166 @@ def __str__(self):
               etree.SubElement(icmp_type_props, "type").text = app[k][prop]
             else:
               pass
-
-    lines = minidom.parseString(
-        etree.tostring(app_entries)).toprettyxml(indent="  ")
```
rdsharma, line 731: Can we add this functionality back in?

rdsharma (changes requested):

capirca/lib/paloaltofw.py
kevinsteves: Added ident of the XML document.

rdsharma (approved):
This looks good, nice work! Will wait until @abhindes reviews as well (for syntax correctness) to pull this in.

Like I mentioned before, we'll have to do something to expose the etree object to tests later on so that we can check the DOM itself rather than the rendered output, but that can come in a later CL if you prefer.

abhindes (approved):
Looks good to me. I loved how easy it was to view the diffs as well.

Great work Kevin!

PR #256, opened 2021-03-31
Place header comments in a new tag object with comments.

kevinsteves:
Then specify that tag name in subsequent rules until
next header.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -278,6 +278,8 @@ class PaloAltoFW(aclgenerator.ACLGenerator):
   _TERM_MAX_LENGTH = 31
   _SUPPORTED_PROTO_NAMES = ["tcp", "udp", "icmp", "icmpv6", "sctp", "igmp"]
   _MAX_RULE_DESCRIPTION_LENGTH = 1024
+  _MAX_TAG_COMMENTS_LENGTH = 1023
+  _TAG_NAME_FORMAT = "{from_zone}_{to_zone}_policy-comment-{num}"
 
   INDENT = "  "
 
@@ -750,10 +752,27 @@ class PaloAltoFW(aclgenerator.ACLGenerator):
     rulebase = etree.SubElement(vsys_entry, "rulebase")
     security = etree.SubElement(rulebase, "security")
     rules = etree.SubElement(security, "rules")
+    tag = etree.Element("tag")
+
+    tag_num = 0
 
     # pytype: disable=key-error
     # pylint: disable=unused-variable
     for (header, pa_rules, filter_options) in self.pafw_policies:
+      tag_name = None
+      if header.comment:
+        comment = " ".join(header.comment).strip()
+        if comment:
+          tag_num += 1
+          # max tag len 127, max zone len 31
+          tag_name = self._TAG_NAME_FORMAT.format(
+            from_zone=filter_options[1], to_zone=filter_options[3],
+            num=tag_num)
+          tag_entry = etree.SubElement(tag, "entry",
+                                       {"name": tag_name})
+          comments = etree.SubElement(tag_entry, "comments")
+          comments.text = comment[:self._MAX_TAG_COMMENTS_LENGTH]
+
       for name, options in pa_rules.items():
         entry = etree.SubElement(rules, "entry", {"name": name})
         if options["description"]:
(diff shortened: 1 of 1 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -750,10 +752,27 @@ def __str__(self):
...
+        if comment:
+          tag_num += 1
+          # max tag len 127, max zone len 31
+          tag_name = self._TAG_NAME_FORMAT.format(
+            from_zone=filter_options[1], to_zone=filter_options[3],
+            num=tag_num)
+          tag_entry = etree.SubElement(tag, "entry",
+                                       {"name": tag_name})
+          comments = etree.SubElement(tag_entry, "comments")
+          comments.text = comment[:self._MAX_TAG_COMMENTS_LENGTH]
```
rdsharma, line 774: This will silently elide the extra comment characters when the comment is too long. We should either error or, ideally, split the comment into multiple comments.

capirca/lib/paloaltofw.py
```diff
@@ -750,10 +752,27 @@ def __str__(self):
...
+    tag_num = 0
 
     # pytype: disable=key-error
     # pylint: disable=unused-variable
     for (header, pa_rules, filter_options) in self.pafw_policies:
+      tag_name = None
+      if header.comment:
+        comment = " ".join(header.comment).strip()
+        if comment:
+          tag_num += 1
```
rdsharma, line 766: Since this is done at the beginning of the clause we'll never print a comment 0. I just want to make sure that's intentional?

rdsharma, line 766: It appears we're incrementing this without checking against a limit. Can PAN handle an unlimited number of comments or do we need to enforce a max?

rdsharma (changes requested):

capirca/lib/paloaltofw.py
abhindes: Since this is based on the number of tags you can have, the limit maybe quite large. I can't find anything published.

capirca/lib/paloaltofw.py
abhindes: I think truncation here is fine. The logic to create a new tag comment and associate multiple tags creates complicated, messy logic. This follows the expected behavior of adhering to comment limits, this shouldn't error out.

capirca/lib/paloaltofw.py
```diff
@@ -750,10 +752,27 @@ def __str__(self):
...
+          # max tag len 127, max zone len 31
+          tag_name = self._TAG_NAME_FORMAT.format(
```
abhindes, line 768: Can we error out here if the tag name exceeds the max length? Or if the tag name is None.

rdsharma:
This change also needs tests.

capirca/lib/paloaltofw.py
kevinsteves: the number in the tag name is intended to start at 1.

kevinsteves: maximum tags is 10,000: https://knowledgebase.paloaltonetworks.com/KCSArticleDetail?id=kA10g000000PO8vCAG&lang=en_US%E2%80%A9

capirca/lib/paloaltofw.py
kevinsteves: tag name len cannot exceed 127 unless zone(s) are > 31, and if zones are > 31 they are invalid.

I don't think format() can return None.

capirca/lib/paloaltofw.py
kevinsteves: policy description can also be truncated.

kevinsteves:
> This change also needs tests.

Plan was to address in subsequent change after exposing ElementTree root and enhancing existing tests to use ElementTree as appropriate.

rdsharma:
> > This change also needs tests.
> 
> Plan was to address in subsequent change after exposing ElementTree root and enhancing existing tests to use ElementTree as appropriate.

Understood, definitely don't need to do the full test conversion yet, but I think we should ideally add a test for the new feature in his PR. Maybe can just add the couple lines needed to expose the root element to tests, and the test for these comments tags, then leave the rest for after?

capirca/lib/paloaltofw.py
abhindes: That sounds good to me. We'll follow up with the rest of the configurable limits.

rdsharma (approved):
Approving based on discussion in meeting - will have followup CLs for testing and limits where needed

capirca/lib/paloaltofw.py
abhindes: Sounds good.

abhindes (approved):

PR #259, opened 2021-04-05
Rework PAN-OS unittests to use ElementTree.

kevinsteves:
Expose ElementTree config root in paloaltofw.PaloAltoFW() as a data
attribute.

Use it in the unittests to replace string matching in the XML (assertIn()
and assertNotIn()).  This results in more accurate verification of tests
using ElementTree's XPath expression support.


rdsharma:
Looks like a file conflicts from previous commit, please fix

rdsharma (approved):

PR #260, opened 2021-04-11
Add tests for PAN-OS header comments.

kevinsteves:
Header comments create a tag object which is referenced in subsequent
security rules.


rdsharma (approved):
Would be nice to see a test that checks that the comment limit is enforced in a later PR, but this LGTM for now

PR #266, opened 2021-04-21
Juniper: Add 'mixed' support

### LINKED ISSUE #257: Support multiple targets per header
Based on https://github.com/google/capirca/wiki/Policy-format#header-section I was under the impression that a header could have multiple targets, for example:
```
header {
  target:: juniper testpol4 inet
  target:: juniper testpol6 inet6
}
term default {
    comment:: "test"
    action:: deny
}
```

But running that only outputs:
```
firewall {
    family inet {
        /*
         ** $Id:$
         ** $Date:$
         ** $Revision:$
         **
         */
        replace: filter testpol4 {
            interface-specific;
            /*
             ** test
             */
            term default {
                then {
                    discard;
                }
            }
        }
    }
}
```
While I would have expected both inet and inet6 filters to be generated.
I'm wondering if it's a regression, bug or was just never supported, but it would be nice to keep policy files as lean as possible.

The workaround is to do:
```
header {
  target:: juniper testpol4 inet
}
#include 'testpol.inc'

header {
  target:: juniper testpol6 inet6
}
#include 'testpol.inc'
```
Which is more complex and creates clutter.

What do you think?

Thanks.

XioNoX:
Global: FilterOptions now returns a list of options if there are multiple targets with the same platform, while keeping backward compatibility by returning the list of options directly when only 1 target is found.

Juniper: Iterate over the options (including the names) and generate the matching filters.

Tests included.

Fixes #257


abhindes:
Would it be more Capirca-native to support "mixed" option for Juniper, rather than creating a new behavior where multiple targets need to be supported (for different address families).

For example, the way the Cisco generator supports "mixed" - https://github.com/google/capirca/blob/master/doc/generator_patterns.md#when-the-platform-does-not-support-mixed-in-a-single-access-list

Instead of having 

```
header {
  target:: juniper testpol4 inet
  target:: juniper testpol6 inet6
}
term default {
    comment:: "test"
    action:: deny
}
```

You would have -
```
header {
  target:: juniper testpol mixed
}
term default {
    comment:: "test"
    action:: deny
}
```

XioNoX:
Using mixed looks cleaner at first. However:
* It doesn't mix v4 and v6 like Cisco but generates 2 filters
* Doesn't let the user chose the filter names (eg. only add the 4 or 6 suffix to the filter name)

If you think that's fine, I can look at updating my PR.

Thanks.

XioNoX:
@abhindes is there anything I can do to move forward on that PR? Thanks

abhindes:
Hi, sorry for the delay in response.

I believe that the right way to do this in a Capirca native method, would be to support "mixed". Could you elaborate a little on 

> It doesn't mix v4 and v6 like Cisco but generates 2 filters

I believe the Cisco generator also creates 2 filters, one for IPv4 and one for IPv6, when "mixed" is used.

> Doesn't let the user chose the filter names (eg. only add the 4 or 6 suffix to the filter name)

I am not sure of the severity of this issue, we could look into a feature request on renaming filter names if required. If you are rendering an option using "mixed", then it is likely the filter names will be semantically related for IPv6 and for IPv6.

XioNoX:
@abhindes it's indeed much cleaner with the "mixed" option!
The CI errors don't seem to be related to this PR though. Please let me know if there is anything to change.

XioNoX:
@abhindes is there anything more I need to do? Thanks

abhindes:
Hi, I'll try and take a look at this (and the other issues/PRs), this week, thanks for your patience! :)

abhindes:
Please take a look at this and update test cases as specified in https://github.com/google/capirca/blob/master/doc/generator_patterns.md#when-the-platform-supports-mixed-in-a-single-access-list. 

While Juniper follows the "The desired approach will be to have Capirca output two filters, one for each address family, for platforms that do not support ‘mixed’.", we still require test cases that validate different flows. 

Please see JuniperSRX/JuniperMSMPC generators if you need some guidance on the tests :)

XioNoX:
> Please take a look at this and update test cases as specified in https://github.com/google/capirca/blob/master/doc/generator_patterns.md#when-the-platform-supports-mixed-in-a-single-access-list.
> 
> While Juniper follows the "The desired approach will be to have Capirca output two filters, one for each address family, for platforms that do not support ‘mixed’.", we still require test cases that validate different flows.
> 
> Please see JuniperSRX/JuniperMSMPC generators if you need some guidance on the tests :)

Thanks! Done. I used junipermsmpc_test.py as inspiration. Let me know if anything is missing.

XioNoX:
Hi @abhindes is there anything more I can do? Thanks

abhindes:
These changes look good. I've resolved the merge conflict, waiting for the tests to run prior to merging.

PR #275, opened 2021-05-19
PAN-OS: Enhancement to handle term source-port keyword.

kevinsteves:
If source-port is specified add it to a service object.  If there is
no destination-port then add range 0-65535 to the service destination
port because it is required field.  The port range is used vs. "any"
due to recent changes which disallow any in the WebUI.

### DIFF
--- a/tests/lib/paloaltofw_test.py
+++ b/tests/lib/paloaltofw_test.py
@@ -885,7 +885,11 @@ term rule-1 {
   destination-port:: NTP
 """
 
-    pol = policy.ParsePolicy(POL % T)
+    definitions = naming.Naming()
+    definitions._ParseLine('NTP = 123/tcp 123/udp', 'services')
+    definitions._ParseLine('DNS = 53/tcp 53/udp', 'services')
+
+    pol = policy.ParsePolicy(POL % T, definitions)
     paloalto = paloaltofw.PaloAltoFW(pol, EXP_INFO)
     output = str(paloalto)
     name = "service-rule-1-udp"
@@ -901,7 +905,7 @@ term rule-1 {
   source-port:: NTP
 """
 
-    pol = policy.ParsePolicy(POL % T)
+    pol = policy.ParsePolicy(POL % T, definitions)
     paloalto = paloaltofw.PaloAltoFW(pol, EXP_INFO)
     output = str(paloalto)
     name = "service-rule-1-udp"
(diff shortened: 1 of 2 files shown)


abhindes (approved):

tests/lib/paloaltofw_test.py
```diff
@@ -869,6 +870,65 @@ def testPanApplication(self):
...
+%s
+  action:: accept
+}"""
+
+    T = """
+  protocol:: udp
+  destination-port:: NTP
+"""
+
+    pol = policy.ParsePolicy(POL % T)
```
abhindes, line 888: This is missing the self.naming argument.

tests/lib/paloaltofw_test.py
```diff
@@ -869,6 +870,65 @@ def testPanApplication(self):
...
+    path = "/entry[@name='%s']/protocol/udp/source-port" % name
+    x = paloalto.config.findtext(PATH_SERVICE + path)
+    self.assertIsNone(x, output)
+
+    T = """
+  protocol:: udp
+  source-port:: NTP
+"""
+
+    pol = policy.ParsePolicy(POL % T)
```
abhindes, line 904: This is missing the self.naming argument.

tests/lib/paloaltofw_test.py
```diff
@@ -869,6 +870,65 @@ def testPanApplication(self):
...
+    x = paloalto.config.findtext(PATH_SERVICE + path)
+    self.assertEqual(x, "123", output)
+
+    T = """
+  protocol:: tcp
+  source-port:: NTP
+  destination-port:: NTP DNS
+"""
+
+    pol = policy.ParsePolicy(POL % T)
```
abhindes, line 921: This is missing the self.naming argument.

abhindes (changes requested):
Please fix the tests.

abhindes:
> > Please fix the tests.
> 
> ParsePolicy() naming definitions are optional, and when not specified
> use the default definitions directory, which I want to use for the
> tests vs. the mock auto-spec naming. Is there an issue with this usage?

The issue with this usage is that it's causing internal tests to fail while merging. I believe this is because our naming directory setup is slightly different. Would it be possible to specify this, like the other tests, so we can merge your changes in?

abhindes (approved):

PR #276, opened 2021-05-26
PAN-OS: Enhancement to handle tcp or udp with no ports specified.

kevinsteves:
If no source-port and destination-port and protocol contains tcp or
udp create (one time) service any-tcp or -any-udp and specify in
service.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -141,12 +141,14 @@ class ServiceMap(object):
   def __init__(self):
     self.entries = {}
 
-  def get_service_name(self, term_name, src_ports, ports, protocol):
+  def get_service_name(self, term_name, src_ports, ports, protocol, prefix=None):
     """Returns service name based on the provided ports and protocol."""
     if (src_ports, ports, protocol) in self.entries:
       return self.entries[(src_ports, ports, protocol)]["name"]
 
-    service_name = "service-%s-%s" % (term_name, protocol)
+    if prefix is None:
+      prefix = "service-"
+    service_name = "%s%s-%s" % (prefix, term_name, protocol)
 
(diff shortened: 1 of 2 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -268,6 +270,21 @@ def pan_ports(ports):
...
+        if p not in ["tcp", "udp"]:
+          logging.warning(
+            "WARNING: Term %s in policy %s>%s contains port-less tcp "
+            "and/or udp with protocol %s.  Move %s to another term.",
+            term.name, self.options["from_zone"][0],
+            self.options["to_zone"][0], p, p)
+          continue
+        ports = pan_ports([("0", "65535")])
+        # use prefix "" to avoid service name clash with term named "any"
+        service_name = service_map.get_service_name("any", (), ports, p, "")
```
abhindes, line 284: This doesn't handle the scenario for when only the source_port is specified, but the destination_port is not, since you're just passing () for src_ports.

abhindes (changes requested):

capirca/lib/paloaltofw.py
kevinsteves: source_port only is handled in the if above: https://github.com/kevinsteves/capirca/blob/br13/capirca/lib/paloaltofw.py#L259

abhindes: Great just saw that this term is correct.

```
<entry name="service-accept-traceroute-source-ports-udp">
              <protocol>
                <udp>
                  <port>0-65535</port>
                  <source-port>33434-33534</source-port>
                </udp>
              </protocol>
            </entry>
```

abhindes (approved):

PR #277, opened 2021-06-23
PAN-OS: Support port-less tcp and udp and other protocols in same term.

kevinsteves:
The term will be expanded into 2 security rules.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -15,6 +15,7 @@
 """Palo Alto Firewall generator."""
 
 import collections
+import copy
 import datetime
 import logging
 import re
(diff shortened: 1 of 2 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -182,20 +183,27 @@ def __init__(self, from_zone, to_zone, terms, service_map):
...
+    self.options = []
+    term = terms.term
+    while term is not None:
+      x, term = self.ModifyOptions(from_zone, to_zone,
+                                   term, service_map)
+      self.options.append(x)
 
-  def ModifyOptions(self, terms, service_map):
+  @staticmethod
+  def ModifyOptions(from_zone, to_zone, term, service_map):
```
rdsharma, line 194: Nit: Should we rename this function now that it doesn't modify the options dict but rather just returns a new one?

rdsharma (commented):
Looks good other than one minor nit

abhindes (approved):

PR #282, opened 2021-07-29
PAN-OS: Handle src any dst any term for inet and inet6 address family.

kevinsteves:
Create address object to match IPv4 addresses for inet; use same
object for inet6 and negate it in the policy.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -238,6 +238,7 @@ class Rule:
       saddr_check = sorted(saddr_check)
       for addr in saddr_check:
         options["source"].append(str(addr))
+    # missing source handled during XML document generation
 
     # DESTINATION-ADDRESS
     if term.destination_address:
@@ -247,6 +248,7 @@ class Rule:
       daddr_check = sorted(daddr_check)
       for addr in daddr_check:
         options["destination"].append(str(addr))
+    # missing destination handled during XML document generation
 
     # ACTION
     if term.action:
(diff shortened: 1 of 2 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -884,6 +882,28 @@ def __str__(self):
...
+          member = etree.SubElement(dest, "member")
+          if options["source"]:
+            member.text = "any"
+          else:
+            af = filter_options[4] if len(filter_options) > 4 else "inet"
+            if af != "mixed":
+              member.text = "any-ipv4"
+              options["source"] = ["any-ipv4"]
+              add_any_ipv4 = True
+              if af == "inet6":
```
abhindes, line 896: A comment here explaining inet6 handling would be helpful.

capirca/lib/paloaltofw.py
```diff
@@ -770,6 +766,8 @@ def _BuildPort(self, ports):
   def __str__(self):
     """Render the output of the PaloAltoFirewall policy into config."""
 
+    add_any_ipv4 = False
```
abhindes, line 769: A comment here indicating that this is used in the case of both inet/inet6 (since it's counterintuitive).

capirca/lib/paloaltofw.py
```diff
@@ -238,8 +238,6 @@ def pan_ports(ports):
       saddr_check = sorted(saddr_check)
       for addr in saddr_check:
         options["source"].append(str(addr))
-    else:
```
abhindes, line 241: Since we usually handle most address processing at once, but in this case we're not, it'd be helpful to leave a comment here (and for Destination-address processing) that the "any" case is handled later.

capirca/lib/paloaltofw.py
```diff
@@ -884,6 +882,28 @@ def __str__(self):
           member = etree.SubElement(from_, "member")
           member.text = x
 
+        dest = etree.SubElement(entry, "destination")
```
abhindes, line 885: We now have destination always rendered prior to the source in a rule (which to me is slightly counterintuitive). 

Is this how the device displays it if I display the xml config? While this doesn't have to be fixed in this commit, I am thinking about the long term adherence to this [requirement](https://github.com/google/capirca/blob/master/doc/generator_patterns.md#syntax-of-the-config-from-the-generator-and-on-device-should-match-when-cryptographically-verified).

abhindes (commented):
The code changes look fine, once the comments are addresses, good to approve :) 

Thanks for making these changes!

abhindes:
Could you please sync your local branch? I see the checks are failing (most likely due to missing commits), and I would like to take a look at the artifacts created (the .xml policy file particularly) before merging this! 

Thanks!

abhindes (approved):

abhindes (approved):

PR #285, opened 2021-08-30
Update config.py

mdakh404:
Hello, config.py has a function [yaml_loader()] that takes a YAML file as its positional argument and deserializes it (loads it) using the yaml module's load function, load generally is an unsafe function, if the user's supplied config file is a malicious YAML file, it will lead to an Arbitrary Code Execution, so I updated the config file to use safe_load rather than load.

Thanks!

### DIFF
--- a/capirca/utils/config.py
+++ b/capirca/utils/config.py
@@ -21,9 +21,9 @@ defaults = {
 def yaml_loader(filename):
   with open(filename, 'r') as f:
     try:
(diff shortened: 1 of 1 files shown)


TheLinuxGuy:
Hi, 
thanks for the PR. I did not notice this before making another commit fixing this same issue, sorry about that.

This was resolved via commit https://github.com/google/capirca/commit/df7711fd23f7f8715865067955eadf9d883bd0e7

PR #286, opened 2021-09-09
Add check for ICMP types used with non-ICMP protocols.

kevinsteves:
Add unit tests for check above, and invalid ICMP type.

Background: this check was previously handled in
aclgenerator.Term.NormalizeIcmpTypes(), and the function call was
removed.  NormalizeIcmpTypes() is not address family "mixed" aware, so
just added the missing check needed.  And the local Term() class is
no longer needed, so remove it.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -65,76 +65,6 @@ class PaloAltoFWBadIcmpTypeError(Error):
   pass
 
 
-class Term(aclgenerator.Term):
-  """Representation of an individual term.
-
-  This is mostly useful for the __str__() method.
-
-  Attributes:
-    obj: a policy.Term object
-    term_type: type of filter to generate, e.g. inet or inet6
-    filter_options: list of remaining target options (zones)
-  """
-
-  ACTIONS = {
-      "accept": "allow",
-      "deny": "deny",
-      "reject": "reset-client",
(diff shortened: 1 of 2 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -687,11 +624,13 @@ def _TranslatePolicy(self, pol, exp_info):
           raise PaloAltoFWUnsupportedProtocolError(
               "protocol %s is not supported" % proto_name)
 
-        # Create Term object with the term, address family, and header
-        # parameters, e.g. to/from zone, and add it to a list of
-        # terms that would form a rule.
-        new_term = Term(term, filter_type, filter_options)
-        new_terms.append(new_term)
+        if (term.icmp_type and term.protocol != ['icmp'] and
```
abhindes, line 627: nit: This will also error out for protocol ['icmp', 'icmpv6'], but the error message is not indicative of this.

From our discussion last week, we said it was okay to not have icmp and icmpv6 in the same term with icm_types. But this error message would be misleading. 

I know that NormalizeIcmpTypes also has the same behavior, but we can start with fixing this.

kevinsteves: Updated with additional check and error message.

abhindes (approved):

abhindes (approved):

PR #287, opened 2021-09-16
Update windows.yml to fix action

abhindes:
The Windows action is using Capirca from the pypi, when it should be installing Capirca from the local filesystem. 

It seems like --no-index not being used causes it to install from pypi, as can be seen in https://github.com/google/capirca/runs/2949891967?check_suite_focus=true#step:7:284


abhindes:
Used for testing, committed via https://github.com/google/capirca/commit/5729be660bd904ec6e68ae5a68832aa236c44853.

Closing this out.

PR #288, opened 2021-10-05
PAN-OS: Add "address objects" header option.

kevinsteves:
address objects: specifies whether custom address objects or
network/mask definitions are used in security policy source and
destination fields
    addr-obj: specifies address groups are used in the security policy
    source and destination fields (default)
    no-addr-obj: specifies network/mask definitions are used in the
    security policy source and destination fields

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -269,6 +269,7 @@ class PaloAltoFW(aclgenerator.ACLGenerator):
   _MAX_RULE_DESCRIPTION_LENGTH = 1024
   _MAX_TAG_COMMENTS_LENGTH = 1023
   _TAG_NAME_FORMAT = "{from_zone}_{to_zone}_policy-comment-{num}"
+  _MAX_RULE_SRC_DST_MEMBERS = 65535
 
   INDENT = "  "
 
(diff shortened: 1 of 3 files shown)


capirca/lib/paloaltofw.py
```diff
@@ -847,9 +894,22 @@ def __str__(self):
...
-            member.text = x
+            if no_addr_obj:
+              for group in address_book_groups_dict[x]:
+                member = etree.SubElement(source, "member")
+                member.text = str(address_book_names_dict[group])
+                max_src_dst += 1
+            else:
+              member = etree.SubElement(source, "member")
+              member.text = x
+              max_src_dst += 1
```
abhindes, line 905: My minor concern with this is that reusing _MAX_RULE_SRC_DST_MEMBERS here makes it seem like the limits for when the address objects are used is the same as when the address objects are not used, but that isn't the case I believe. 

I cannot recall a known limit that we discussed to modify this to, so looking for some input here. 

From looking at product docs, I see that limits are:

Address objects | 30,000 
Address groups | 15,000 
Members per address group | 2,500 <--- what we probably need?

abhindes: Oops, if address objects are used, this will be a single address group entry in the rule. Should be fine then. I have a separate concern about the address group limits that we can probably resolve outside this PR. Resolving.

abhindes (approved):

Issue #284, opened 2021-08-27
Paloalto generator is adding application "any" to list of applications even if the application already contains a more specific application


hkam40k:
Hello,

https://github.com/google/capirca/blob/master/capirca/lib/paloaltofw.py#L309
```
    if term.protocol:
      # Add application "any" to all terms, unless ICMP/ICMPv6
      for proto_name in term.protocol:
        if proto_name in ["icmp", "icmpv6"]:
          continue
        elif proto_name in ["igmp", "sctp", "gre"]:
          if proto_name not in options["application"]:
            options["application"].append(proto_name)
        elif proto_name in ["tcp", "udp"]:
          if "any" not in options["application"]:  # <----------
            options["application"].append("any")
        else:
          pass
```
The above seems a bug to me, and can result in a case where application list is for example "web-browsing, any". It would make more sense if the "any" is added only if the list is empty, but that would still be redundant with https://github.com/google/capirca/blob/master/capirca/lib/paloaltofw.py#L934 

The block seems to explicitly handle every item included in `_SUPPORTED_PROTO_NAMES`, but the whole block could possibly be reduced to something like:
```
    if term.protocol:
      # Add certain protocol names as application in the application list, if missing.
      for proto_name in term.protocol:
        if proto_name in ["igmp", "sctp", "gre"]:
          if proto_name not in self.options["application"]:
            self.options["application"].append(proto_name)
```
If you agree that this is a bug, and on the above suggested solution, then I can submit a PR to resolve it.

hkam40k:
Since PANOS 10.0.7, mixing application "any" with a more specific application will raise a blocking error during configuration validation, so this should definitely be considered a bug.

abhindes:
Hi,

Yes, this is definitely a bug. Thanks for raising this issue! 

@kevinsteves is working on #274 and to ensure we don't cause a regression, we will take a look at this as soon as the fix for that is proposed. 

Thanks for your comment on #274, please not that #274 may not disable the usage of pan-application and services (tcp/udp), but will make sure the safe thing occurs (that the usage of tcp/udp/icmp and pan-application ssl in the same rule does not cause a regression). Once #274 is investigated, we will post on how to proceed with that. 

If #284 still makes sense as is once that is complete, you are welcome to send a PR, thanks! :)

abhindes:
Hi,

Just a gentle ping on this, thanks for pointing this issue out, feel free to send a CL anytime. 

[...]

PR #290, opened 2021-11-09
Define and document behavior for pan-application usage.

kevinsteves:
pan-application:: paloalto target only. Specify applications for the
security policy which can be predefined applications
(https://applipedia.paloaltonetworks.com/) and custom application
objects.

  Security Policy Service Setting

    When no protocol is specified in the term, the service will be
    application-default.

    When protocol is tcp or udp, and no source-port or
    destination-port is specified, the service will be custom service
    objects for the protocols and all ports (0-65535).

    When protocol is tcp or udp, and a source-port or destination-port
    is specified, the service will be custom service objects for the
    protocols and ports.

    pan-application can only be used when no protocol is specified in
    the term, or the protocols tcp and udp.

### DIFF
--- a/capirca/lib/paloaltofw.py
+++ b/capirca/lib/paloaltofw.py
@@ -431,6 +431,14 @@ class PaloAltoFW(aclgenerator.ACLGenerator):
                                              term.name)
         term_dup_check.add(term.name)
 
+        services = {"tcp", "udp"} & set(term.protocol)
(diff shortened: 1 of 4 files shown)


abhindes (approved):
This still requires #284 to be fixed to be perfect, but otherwise looks great. Thanks for all the work!

Issue #295, opened 2022-01-24
Support for IOSXR IPv6 Object Group ACLs


PhillSimonds:
The following options are currently available for configuring a Cisco ACL on iosxr:

```
target:: ciscoxr [filter name] {extended|standard|object-group|inet6|mixed} {dsmo}
```

Cisco XR devices support both ipv4 and ipv6 object-group based ACLs. The `object-group` keyword causes capirca to generate an IPv4 object-group ACL.

The following inputs...

```
# NETWORK.net
DEFAULT_ROUTE_V4 = 0.0.0.0/0

RFC_1918 = 192.168.0.0/16
           172.12.0.0/12
           10.0.0.0/8

IPV6_BOGON = fc00::/7
             fe80::/10
             fec0::/10
             ff00::/8
             ::/96
             ::/128
             ::1/128
             ::ffff:0.0.0.0/96
             64:ff9b::/96

# POLICY.pol
header {
  target:: ciscoxr INBOUND_EDGE_V4 object-group
}

term DENY_RFC1918 {
    source-address:: RFC_1918
    action:: deny
}

header {
    target:: ciscoxr INBOUND_EDGE_v6 object-group
}

term DENY_IPV6_BOGONS {
    source-address:: IPV6_BOGON
    action:: deny
}
```

Yield this output

```
# Generated ACL
object-group network ipv6 IPV6_BOGON
 ::/96
 ::ffff:0:0/96
 64:ff9b::/96
 fc00::/7
 fe80::/9
 ff00::/8
exit

object-group network ipv4 RFC_1918
 10.0.0.0/8
 172.0.0.0/12
 192.168.0.0/16

no ipv4 access-list INBOUND_EDGE_V4
ipv4 access-list INBOUND_EDGE_V4
 remark $Id:$


 remark DENY_RFC1918
 deny ip net-group RFC_1918 any

exit

no ipv4 access-list INBOUND_EDGE_v6
ipv4 access-list INBOUND_EDGE_v6
 remark $Id:$


 remark DENY_IPV6_BOGONS
 deny ip net-group IPV6_BOGON any

exit
```

We have a use case such that we require IPv6 object-group based ACLs. It would be nice if we could specify an additional keyword of `object-group-inet6` to render IPv6 object groups. E.g. we could rename the above POLICY.pol declaration to the following:

```
# POLICY.pol
header {
    target:: ciscoxr INBOUND_EDGE_v6 object-group-inet6
}

term DENY_IPV6_BOGONS {
[...]

PR #297, opened 2022-02-11
Adds object-group support for IPv6 to Cisco generator

abates:
This PR fixes issue #295 by adding the "object-group-inet6" argument for the Cisco generator.  This change allows policies to be generated that reference object groups for both ip and ip6 access lists.

### DIFF
--- a/tests/lib/cisco_test.py
+++ b/tests/lib/cisco_test.py
@@ -50,10 +50,16 @@ header {
   target:: cisco 50 standard
 }
 """
-GOOD_OBJGRP_HEADER = """
+GOOD_OBJGRP_HEADER_1 = """
 header {
   comment:: "obj group header test"
   target:: cisco objgroupheader object-group
+}
+"""
+GOOD_OBJGRP_HEADER_2 = """
+header {
+  comment:: "obj group header test"
+  target:: cisco objgroupheader object-group-inet6
 }
 """
 GOOD_INET6_HEADER = """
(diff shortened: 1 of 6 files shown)


tests/lib/cisco_test.py
```diff
@@ -657,6 +662,47 @@ def testObjectGroup(self):
...
+    ip_grp.append('exit')
+    port_grp1 = ['object-group port 80-80']
+    port_grp1.append(' eq 80')
+    port_grp1.append('exit')
+    port_grp2 = ['object-group port 1024-65535']
+    port_grp2.append(' range 1024 65535')
+    port_grp2.append('exit')
+
+    self.naming.GetNetAddr.return_value = [
+        nacaddr.IP('2001::3/128', token='SOME_HOST')]
```
abhindes, line 677: Could you have this also return an IPv4 address, and then we can add a negative test (assertNotIn) for that IPv4 address to ensure IPv4 addresses are not rendered in this object group.

abhindes (changes requested):

tests/lib/cisco_test.py
abates: I'm happy to update the tests, but I'm not exactly sure what you want me to test here. If `GetNetAddr` returns a list including an IPv4 address, then an IPv4 object-group is still created in the policy output. However, that isn't something that was updated in this PR.

I may be misunderstanding, because it seems like all network object groups are created, regardless of address family. See [here](https://github.com/google/capirca/blob/f844501277d057c12a79148839f595762763ebdf/capirca/lib/cisco.py#L250-L254).

abhindes:
Hi, just following up on this.

abhindes:
hi abates, sorry for the trouble, but I think another internal change pushed right now is causing another merge conflict. If you could please re-sync once more time, thank you!

PR #300, opened 2022-03-04
Adds icmp type and code to Cisco object-group output

### LINKED ISSUE #298: Cisco object-group ACLs don't properly render ICMP type and code
Policy terms with an icmp type and code are not correctly rendered when using the `Cisco` generator and the `object-group` argument.  The rendered lines are missing the icmp-type and code.

For example, using the following policy:
```
# NETWORK.net
header {
    target:: ciscoxr ICMP_EXAMPLE object-group
}

term PERMIT_ICMP_ECHO_REPLY {
    protocol:: icmp
    action:: accept
    icmp-type:: echo-reply
}
```

`aclgen` renders the following:

```
! $Id:$
! $Date:$
! $Revision:$
no ip access-list extended ICMP_EXAMPLE
ip access-list extended ICMP_EXAMPLE
 remark $Id:$


 remark PERMIT_ICMP_ECHO_REPLY
 permit icmp any any

exit
```

Note the missing icmp-type following `permit icmp any any`. The expected output is:
```
! $Id:$
! $Date:$
! $Revision:$
no ip access-list extended ICMP_EXAMPLE
ip access-list extended ICMP_EXAMPLE
 remark $Id:$


 remark PERMIT_ICMP_ECHO_REPLY
 permit icmp any any 0

exit
```

abates:
icmp-type and icmp-code were not being rendered in the
ACL when using the object-group feature of the Cisco
renderer. This commit adds icmp type and code to the
output.

Fixes #298

### DIFF
--- a/capirca/lib/cisco.py
+++ b/capirca/lib/cisco.py
@@ -837,6 +837,15 @@ class ObjectGroupTerm(Term):
                   else self.PROTO_MAP.get(proto)
                   for proto in self.term.protocol]
 
+    # icmp lookups
+    icmp_types = ['']
+    if self.term.icmp_type:
+      icmp_types = self.NormalizeIcmpTypes(self.term.icmp_type,
+                                           self.term.protocol, self.af)
+    icmp_codes = ['']
+    if self.term.icmp_code:
+      icmp_codes = self.term.icmp_code
+
     # addresses
     source_address = self.term.source_address
     if not self.term.source_address:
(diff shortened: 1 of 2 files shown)


capirca/lib/cisco.py
```diff
@@ -859,14 +868,24 @@ def __str__(self):
...
           for dport in destination_port:
             for proto in protocol:
-              ret_str.append(
-                  self._TermletToStr(_ACTION_TABLE.get(str(
-                      self.term.action[0])), proto, saddr, sport, daddr, dport))
+              # cisconx uses icmp for both ipv4 and ipv6
+              if self.platform == 'cisconx':
+                if self.af == 6:
+                  proto = 'icmp' if proto == 'icmpv6' else proto
+              for icmp_type in icmp_types:
```
abhindes, line 875: Rather than looping here, making an explicit decision that this is for "icmp" handling only would make this correct, and readable. 

In the case of icmp/icmpv6, this would handle it with the way you have, with self._TermletToStr(_ACTION_TABLE.get(str(
                          self.term.action[0])), proto, saddr, sport, daddr, dport, icmp_type, icmp_code)

Else, it would handle it the way it originally was. 
self._TermletToStr(_ACTION_TABLE.get(str(
                      self.term.action[0])), proto, saddr, sport, daddr, dport))

abhindes (changes requested):
Thanks for sending this PR!

tests/lib/cisco_test.py
```diff
@@ -656,6 +673,21 @@ def testObjectGroup(self):
     self.naming.GetNetAddr.assert_has_calls([mock.call('SOME_HOST'),
                                              mock.call('SOME_HOST')])
     self.naming.GetServiceByProto.assert_called_once_with('HTTP', 'tcp')
+  
+  def testObjectGroupIcmp(self):
```
abhindes, line 677: Could you also please add a testcase for icmpv6 here, thanks!

PR #307, opened 2022-06-04
New flag: --[no]apply_groups to enable/disable apply-groups creation for Juniper MS-MPC

eimann:
Currently capirca generates Juniper MS-MPC as apply-groups. The other way to use those rules is to directly embed them in the services stateful-firewall hierachy.
This PR introduces a switch apply_groups which changes this behaviour. It is set to True by default to not change default behaviour.


eimann:
We are not using capirca anymore, so won't invest time in maintaining this feature anymore.

PR #308, opened 2022-06-15
Add support for 'decapsulate' on Juniper platforms

BarbarossaTM:
This adds support for `decapsulate` actions on Juniper, for example `decapsulate mpls-in-udp` which is available on (at least some) MXes and QFXes.

Trying to figure out if CLA is already in place, opening this already to gather feedback :-)

### DIFF
--- a/capirca/lib/juniper.py
+++ b/capirca/lib/juniper.py
@@ -140,6 +140,7 @@ class Term(aclgenerator.Term):
              'next': 'next term',
              'reject-with-tcp-rst': 'reject tcp-reset',
              'encapsulate': 'encapsulate',
+             'decapsulate': 'decapsulate',
              'port-mirror': 'port-mirror'}
 
   # the following lookup table is used to map between the various types of
@@ -566,6 +567,8 @@ class Term(aclgenerator.Term):
       unique_actions.update(self.term.action)
     if self.term.encapsulate:
       unique_actions.add('encapsulate')
+    if self.term.decapsulate:
+      unique_actions.add('decapsulate')
     if len(unique_actions) <= 1:
       for action in [
           self.term.logging, self.term.routing_instance, self.term.counter,
@@ -609,6 +612,10 @@ class Term(aclgenerator.Term):
       elif current_action == 'encapsulate':
         config.Append('then {')
         config.Append('encapsulate %s;' % str(self.term.encapsulate))
+        config.Append('}')
+      elif current_action == 'decapsulate':
+        config.Append('then {')
+        config.Append('decapsulate %s;' % str(self.term.decapsulate))
         config.Append('}')
       else:
         config.Append('then %s;' % current_action)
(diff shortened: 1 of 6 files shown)


BarbarossaTM:
Hey folks, is any additional step from my end required to review this? :-)

capirca/lib/juniper.py
```diff
@@ -655,6 +662,8 @@ def __str__(self):
           config.Append('next-ip6 %s;' % str(self.term.next_ip[0]))
       if self.term.encapsulate:
         config.Append('encapsulate %s;' % str(self.term.encapsulate))
+      if self.term.decapsulate:
```
abhindes, line 665: We probably want this in an elif block to ensure only one or the other is used. 

"encapsulate" and "decapsulate" are terminating actions. In a given firewall filter term, we cannot have 2 different terminating actions.

So we'd also want to error out somewhere if both encapsulate and decapsulate terminating actions are encountered (and an associated test for this). 

Thanks!

abhindes (changes requested):

abhindes:
Hi, thanks for raising this PR. I've left a minor comment, otherwise these changes look good to merge! Thanks!

rdsharma:
This branch also needs a rebase to fix merge conflicts, can you please do that when you update?

BarbarossaTM:
Hey folks, just came back from vacation :-) Thanks for the feedback, will rebase right away. :)

capirca/lib/juniper.py
BarbarossaTM: Shouldn't this be caught by the following? Happy to move to `elif` if you prefer :-)
```python
 568     if self.term.encapsulate:
 569       unique_actions.add('encapsulate')
 570     if self.term.decapsulate:
 571       unique_actions.add('decapsulate')
```

BarbarossaTM: Had another look at this and added and explicit test for `ecanpsulate`+ `decapsulate` together, which already was caught by the check above :-)

BarbarossaTM:
Update PR again with another explicit test for `encapsulate` + `decapsulate`. I believe we should be good now :-)

BarbarossaTM:
Hey @abhindes & @rdsharma, any update on this? :-)

abhindes:
> Hey @abhindes & @rdsharma, any update on this? :-)
Working on it!

abhindes:
> > Hey @abhindes & @rdsharma, any update on this? :-)
> > Working on it!
Could you try rebase this once more? We're not seeing this issue at head and want to rule out a rebasing issue. Thanks!

tests/lib/policy_test.py
```diff
@@ -437,6 +437,11 @@
   source-zone:: zone1 zone2
   destination-zone:: zone1 zone2
   action:: accept
+"""
```
rdsharma, line 440: It looks like the test failures are actually due to a close bracket getting effectively deleted here. Please fix this and rebase for us to merge.

rdsharma (changes requested):

tests/lib/policy_test.py
BarbarossaTM: Damn, didn't spot this, fixed :-)

PR #325, opened 2022-09-10
Add support for ttl-except to the Juniper platform

nward:
Add support for the `ttl-except` match condition to the Juniper platform.

Have added tests, and all passing fine.

Generates valid config.


nward:
I'm unable to sign the CLA as my org (SearchLight) because https://cla.developers.google.com/clas doesn't let me switch to another Google account I'm logged in as (I have.. 9 or so), so it does not recognise that I am the owner of the group.

I have attempted to switch accounts on https://developers.google.com and it does not take effect on the CLA URL above.

I have signed the CLA personally, and I am a Director of the org.

¯\_(ツ)_/¯

PR #338, opened 2023-02-27
initial commit of NSXT integration

g-nikoloff:
feature branch for pull request in capirca


.pre-commit-config.yaml
```diff
@@ -1,8 +1,4 @@
 repos:
-  - repo: https://gitlab.com/pycqa/flake8
```
ivucica, line 2: Please roll back these changes, or explain why they should be upstreamed.

capirca/aclgen.py
```diff
@@ -13,7 +13,6 @@
 # limitations under the License.
 #
 """Renders policy source files into actual Access Control Lists."""
-
```
ivucica, line 16: Please minimize the amount of formatting changes like this one.

capirca/aclgen.py
```diff
@@ -183,6 +183,7 @@ def RenderFile(base_directory: str, input_file: pathlib.Path,
   msmpc = False
   spd = False
   nsx = False
+  nsxt1 = False
```
ivucica, line 186: Why `nsxt1` instead of `nsxt`? 

Is this because the module is called `nsxt` and you don't want to overwrite the global variable referencing the `nsxt.py` module? Can we call this something different, e.g. `nsxt_pol`?

doc/generators/nsxt.md
```diff
@@ -0,0 +1,54 @@
...
+```
+  * _section_name_: specifies the name of the dfw rule all terms in this header apply to. [mandatory field]
+  * _inet_: specifies the output should be for IPv4 only filters. This is the default format.
+  * _inet6_: specifies the output be for IPv6 only filters.
+  * _mixed_: specifies that the resulting filter should render both IPv4 and IPv6 addresses.
+  * _sectionId_: specifies the Id for the section [optional]
+  * _securitygroup_: specifies that the appliedTo should be security group [optional]
+  * _securitygroupId_: specifies the Id of the security group [mandatory if securitygroup is given]
+(Required keywords option and verbatim are not supported in NSX)
+## Term Format
```
ivucica, line 31: Shouldn't there be a blank line before and after a `##` paragraph in markdown?

doc/generators/nsxt.md
```diff
@@ -0,0 +1,54 @@
...
+* _option::_ See platforms supported Options section.
+* _platform::_ one or more target platforms for which this term should ONLY be rendered.
+*_platform-exclude:: one or more target platforms for which this term should NEVER be rendered.
+* _protocol::_ the network protocols this term will match, such as tcp, udp, icmp, or a numeric value.
+* _source-address::_ one or more source address tokens.
+* _source-exclude::_ exclude one or more address tokens from the specified source-address.
+* _source-port::_ one or more service definition tokens.
+* _verbatim::_ this specifies that the text enclosed within quotes should be rendered into the output without interpretation or modification.  This is sometimes used as a temporary workaround while new required features are being added.
+## Sub Tokens
+### Actions
```
ivucica, line 50: Same as above: I suspect there should be blank lines around `##` and `###` header paragraphs.

tests/lib/nsxt_functtest.py
```diff
@@ -0,0 +1,117 @@
+# Copyright 2015 The Capirca Project Authors All Rights Reserved.
```
ivucica, line 1: Should the years be updated?

tests/lib/nsxt_functtest.py
```diff
@@ -0,0 +1,117 @@
...
+    pass
+
+  def test_nsxt_policy(self):
+    pol = policy.ParsePolicy(nsxt_mocktest.POLICY, self.defs)
+    exp_info = 2
+    nsx = copy.deepcopy(pol)
+    fw = nsxt.Nsxt(nsx, exp_info)
+    output = str(fw)
+
+    # parse the xml
```
ivucica, line 58: This entire test seems strange. Why is it processing bits of XML if what we're generating is JSON? Did this test execute and pass for you?

tests/lib/nsxt_unittest.py
```diff
@@ -0,0 +1,201 @@
...
+      rule_str = nsxt.Term.__str__(nsxt_term)
+    # parse xml rule and check if the values are correct
+    root = ET.fromstring(rule_str)
+    # check name and action
+    self.assertEqual(root.find('name').text, 'allow-ntp-request')
+    self.assertEqual(root.find('action').text, 'allow')
+
+    # check source address
+    exp_sourceaddr = ['10.0.0.1', '10.0.0.2']
+    for destination in root.findall('./sources/source'):
```
ivucica, line 90: This seems like it involves XML. Has this test been updated for the new JSON output?

tests/lib/nsxt_unittest.py
```diff
@@ -0,0 +1,201 @@
...
+    """Test for Term._ServiceToStr."""
+
+    proto = 6
+    icmp_types = []
+    dports = [(1024, 65535)]
+    spots = [(123, 123)]
+    nsxt_term = nsxt.Term(nsxt_mocktest.INET_TERM, 'inet')
+    service = nsxt_term._ServiceToString(proto, spots, dports, icmp_types)
+    self.assertEqual(
+        service, '<service><protocol>6</protocol><sourcePort>123'
```
ivucica, line 71: The expected generated service term should be JSON, not XML, right?

capirca/aclgen.py
```diff
@@ -46,6 +45,7 @@
 from capirca.lib import naming
 from capirca.lib import nftables
 from capirca.lib import nsxv
+from capirca.lib import nsxt
```
ivucica, line 48: The `nsxt.py` module itself, the most important aspect of this PR, is missing from this pull request. Please make sure you include everything required in the pull request.

ivucica (changes requested):

.pre-commit-config.yaml
ivucica: @g-nikoloff tells me he is having trouble with `pre-commit.com`'s hooks accessing https://gitlab.com/pycqa/flake8. 

Is this hook correct? Did this move to [github.com/PyCQA/flake8](https://github.com/PyCQA/flake8)?

ivucica: I have created https://github.com/google/capirca/pull/339 to fix this hook.

I suggest rebasing on top of my commit.

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,339 @@
...
+        source_address.append(str(i))
+
+    destination_address = 'ANY'
+    if self.term.destination_address:
+      destination_address = []
+      for i in self.term.destination_address:
+        destination_address.append(str(i))
+
+    services = 'ANY'
+    # if self.term.services:
```
ivucica, line 150: - Is commenting this out correct?
- What side effects does commenting this out have?

Assuming commenting out is correct, the code should be deleted, not commented out.

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,339 @@
...
+      "notes": notes,
+      "direction": direction,
+      "tag": tag,
+      "ip_protocol": ip_protocol
+    }
+
+    # ret_str = []
+
+    # ret_lines = []
+    # ret_lines.append('%s' % (rule))
```
ivucica, line 197: Let's not leave commented-out code around.

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,339 @@
...
+
+    # remove any trailing spaces and replace multiple spaces with singles
+    # stripped_ret_lines = [re.sub(r'\s+', ' ', x).rstrip() for x in ret_lines]
+    # ret_str.extend(stripped_ret_lines)
+
+    print(rule)
```
ivucica, line 203: Let's not leave a debug print to stdout when submitting the code. Check for correctness should be done using a test.

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,339 @@
...
+    #   ip_protocol = self.term.ip_protocol
+
+    rule =  {
+      "action": action,
+      "resource_type": "Rule",
+      "display_name": name,
+      "source_groups": source_address,
+      "destination_groups": destination_address,
+      "services": [
+          services
```
ivucica, line 180: Won't this _always_ use `"ANY"` as the only service in the list? Is that correct?

Same for `profiles` field, etc.

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,339 @@
...
+      "profiles": [
+          profiles
+      ],
+      "scope": [
+          scope
+      ],
+      "notes": notes,
+      "direction": direction,
+      "tag": tag,
+      "ip_protocol": ip_protocol
```
ivucica, line 191: Won't this simply use `"ANY"` as the `ip_protocol`, ignoring that it may be `inet`, `inet6` and `mixed`?

tests/lib/nsxt_test.py
```diff
@@ -0,0 +1,930 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
```
ivucica, line 1: Year. Also please use "The Capirca Project" to match the other files

tests/lib/nsxt_test.py
```diff
@@ -0,0 +1,930 @@
...
+  def testServiceToStr(self):
+    """Test for Term._ServiceToStr."""
+
+    proto = 6
+    icmp_types = []
+    dports = [(1024, 65535)]
+    spots = [(123, 123)]
+    nsxt_term = nsxt.Term(INET_TERM, 'inet')
+    service = nsxt_term._ServiceToString(proto, spots, dports, icmp_types)
+    self.assertEqual(service, '<service><protocol>6</protocol><sourcePort>'
```
ivucica, line 425: More XML (although you have `@unittest.skip` above; would it not make sense to simply delete the test then, if it'll unconditionally be skipped? Is it correct to skip it?)

tests/lib/nsxt_unittest.py
```diff
@@ -0,0 +1,201 @@
+# Copyright 2015 The Capirca Project Authors All Rights Reserved.
```
ivucica, line 1: Year

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,339 @@
+# Copyright 2015 The Capirca Project Authors All Rights Reserved.
```
ivucica, line 1: Year

ivucica (changes requested):

.pre-commit-config.yaml
g-nikoloff: I also changed it in my commit, so there wont be any conflict and we can merge it without any problems.

capirca/aclgen.py
g-nikoloff: Yes, having this as nsxt was overwriting the global and code didn't work. I could rename it to nsxt_pol, my mind just blanked at the time for a good variable name.

doc/generators/nsxt.md
g-nikoloff: Not in this case - a blank line above will move it at the same level as Nsxt and a blank line below moves the content after at the same level as ##Term Format and I want it to be a subsection of it.

doc/generators/nsxt.md
g-nikoloff: Same thing as above - ### Actions are a sub-section of ## Sub Tokens

tests/lib/nsxt_test.py
g-nikoloff: I believe all XML references were removed with the last commit.

capirca/aclgen.py
g-nikoloff: Done

tests/lib/nsxt_unittest.py
g-nikoloff: All the XML content was removed with the last commit, it was a bit of a mess moving from one repo to another.

tests/lib/nsxt_unittest.py
g-nikoloff: All the XML content was removed with the last commit, it was a bit of a mess moving from one repo to another.

tests/lib/nsxt_functtest.py
g-nikoloff: All the XML content was removed with the last commit, it was a bit of a mess moving from one repo to another.

capirca/lib/nsxt.py
g-nikoloff: Added protocol to be used instead.

capirca/lib/nsxt.py
g-nikoloff: left them with default values as those are mandatory for the API call and removed commented code

capirca/lib/nsxt.py
g-nikoloff: removed commented code and left with default values as those are mandatory fields in the API call

ivucica:
Revert won't do the trick; the commit is still part of the history.

You need to actually edit the history and then do a `git push --force`. (Be careful with this, and keep a backup of the full repo if you haven't done this stuff before. Just duplicate the directory locally, or at least the `.git` folder.)

ivucica:
After taking a backup, try this:

```
git rebase -i ebd6b88  # note this is the hash of your first commit
```

Then edit the text file that opens, using `fixup` on the commit you want to make disappear into its parent (the "line" above).

There's a way to just edit authors as well, but that's even trickier.

ivucica:
Something went wrong with the rebase and the PR now includes some previously committed code. This could be because `Fix bug in nftables generator` got rebased on top of mainline branch, instead of the initial commit ebd6b88ab5938dc58d089b2192ccf093fbb07a37 getting rebased.

This will solve both the potential merge conflicts and the CLA issue. 

- When merging, we'll likely want to squash this PR. Maybe we can improve the PR description so the squashed merge looks cleaner as well.
- I'll also try to do another review of the contribution itself.

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,320 @@
...
+        source_address.append(str(i))
+
+    destination_address = 'ANY'
+    if self.term.destination_address:
+      destination_address = []
+      for i in self.term.destination_address:
+        destination_address.append(str(i))
+
+# Intentionally left with defaul values as these fields are mandatory in the API. If they need to be configured these should be extended to br read from the policy files.
+    services = 'ANY'
```
ivucica, line 152: In NSX-V, the XML for a service is output like this:

```python
          services.append(self._ServiceToString(proto,
                                                source_port,
                                                destination_port,
                                                icmp_types))
```

The NSX-T code does not seem to be using `source_port`, `destination_port`, `icmp_types`?

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,320 @@
...
+
+    profiles = 'ANY'
+
+    scope = 'ANY'
+
+    direction = 'IN_OUT'
```
ivucica, line 158: Probably changes depending on whether src vs dest address are specified?

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,320 @@
...
+
+    notes = ''
+    if self.term.comment:
+      notes = self.term.comment
+
+    action = 'ALLOW'
+    if self.term.action:
+      action = self.term.action
+
+    source_address = 'ANY'
```
ivucica, line 139: At what point do security groups get passed in?

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,320 @@
...
+            applied_to = filter_options[index + 1]
+            break
+          else:
+            raise UnsupportedNsxtAccessListError(
+                'Security Group Id is not provided for %s' % (self._PLATFORM))
+
+    self._FILTER_OPTIONS_DICT['section_name'] = section_name
+    self._FILTER_OPTIONS_DICT['filter_type'] = filter_type
+    self._FILTER_OPTIONS_DICT['section_id'] = section_id
+    self._FILTER_OPTIONS_DICT['applied_to'] = applied_to
```
ivucica, line 300: Looks like applied_to -- the ID of the security group -- does not get used anywhere?

tests/lib/nsxt_mocktest.py
```diff
@@ -0,0 +1,140 @@
...
+  """
+
+INET6_TERM = """\
+  term test-icmpv6 {
+    protocol:: icmpv6
+    icmp-type:: echo-request echo-reply
+    action:: accept
+  }
+  """
+
```
ivucica, line 34: Some (or all) of the tests should specify a security group in inputs and outputs, so that we can be sure security group gets used.

tests/lib/nsxt_unittest.py
```diff
@@ -0,0 +1,137 @@
...
+        help='definitions directory',
+        default='def')
+    (FLAGS, args) = parser.parse_args([])
+    self.defs = naming.Naming(FLAGS.definitions)
+
+  def tearDown(self):
+    super().setUp()
+    pass
+
+  def runTest(self):
```
ivucica, line 47: Where does the need for `runTest` come from? absl tests should not require this.

tests/lib/nsxt_unittest.py
```diff
@@ -0,0 +1,137 @@
...
+      self.assertEqual(len(terms), 1)
+
+  def test_nsxt_str(self):
+    """Test for Nsxt._str_."""
+    # exp_info default is 2
+    exp_info = 2
+    pol = policy.ParsePolicy(nsxt_mocktest.MIXED_FILTER, self.defs, False)
+    target = nsxt.Nsxt(pol, exp_info)
+
+    # parse the xml and check the values
```
ivucica, line 114: not xml

tests/lib/nsxt_unittest.py
```diff
@@ -0,0 +1,137 @@
...
+        self.fail('IPv4Address not found in test_nsxt_str()')
+
+    # check protocol
+    protocol = rule["ip_protocol"]
+    self.assertEqual(protocol, ['udp'])
+
+    # check notes
+    notes = rule["notes"]
+    self.assertEqual(notes[0], 'Allow name resolution using honestdns.')
+
```
ivucica, line 135: May I suggest adding additional tests that validate the EXACT JSON generated?

For example:

- update nsxt_mocktest.py to also include the expected output just under the expected input
- on the bottom of nsxt_mocktest.py, define a new list of 2-item tuples e.g. `ALL_VALUES`, where the first item is the expected inputs and the second item is the output JSON
- write an additional test in one of the files that iterates over every tuple, and checks the equality of JSON from the invocation of `nsxt.Nsxt(first_item)` to the `second_item` of the tuple

This is important because it will allow *immediate* visual inspection of whether everything is tested and what the *real* outputs are expected for inputs.

Essentially, let's come up with 'golden outputs' for inputs.

jgangel:
> After taking a backup, try this:
> 
> ```
> git rebase -i ebd6b88  # note this is the hash of your first commit
> ```
> 
> Then edit the text file that opens, using `fixup` on the commit you want to make disappear into its parent (the "line" above).
> 
> There's a way to just edit authors as well, but that's even trickier.
> 
> Please note that, without trying this out locally, I am not in fact sure this is the correct incantation to get the list of all commits on PR. This is a finicky operation and I advise care and careful examination of `git log` output before doing a `git push -f`.

Hi @ivucica, just to be sure, do you need us to squash our commits into a single one for clearer understanding? That would mean to force push it as well.

capirca/aclgen.py
```diff
@@ -442,24 +450,23 @@ def RenderFile(base_directory: str, input_file: pathlib.Path,
...
-      aclgenerator.Error,
-      aruba.Error,
-      nftables.Error,
-      gce.Error,
-      gce_vpc_tf.Error,
-      cloudarmor.Error,
-      k8s.Error) as e:
-    raise ACLGeneratorError('Error generating target ACL for %s:\n%s' %
-                            (input_file, e))
+    juniper.Error,
```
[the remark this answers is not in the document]
jgangel: This suggestion conflicts with the flake8 pre-commit hook

tests/lib/nsxt_unittest.py
jgangel: Is it reasonable to replace the current assertions with the full JSON validation proposed? To not have duplicated test scenarios

capirca/lib/nsxt.py
```diff
@@ -0,0 +1,320 @@
...
+    if unsupported_keywords:
+      logging.warning('WARNING: The keywords %s in Term %s are not supported '
+                      'in Nsxt ', unsupported_keywords, self.term.name)
+
+    name = self.term.name
+
+    notes = ''
+    if self.term.comment:
+      notes = self.term.comment
+
```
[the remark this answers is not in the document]
ivucica: This is done (`"logged": bool(self.term.logging),`).

capirca/lib/nsxt.py
```diff
@@ -137,69 +189,55 @@ def __str__(self):
...
-      "services": [
-          services
-      ],
-      "profiles": [
-          profiles
-      ],
-      "scope": [
-          scope
-      ],
+      "services": ["ANY"],
```
ivucica, line 216: Please document that `"services": ["ANY"],` means that `service_entries` gets used. (I'm referencing a comment from another contribution which doesn't have a PR open: https://github.com/google/capirca/compare/master...jcatrouillet:capirca:master#diff-c11ad0bd48e47a1421118f82ccb0b4bdc5b943a36389e9c0f745e3f3a850d879R169-R170)

capirca/lib/nsxt.py
```diff
@@ -137,69 +189,55 @@ def __str__(self):
...
+      "profiles": ["ANY"],
+      "scope": ["ANY"],
       "logged": bool(self.term.logging),
       "notes": notes,
```
ivucica, line 220: `"notes"` seems ok. However, what is the difference between `"notes"` and `"description"`? (Referencing another contribution that does not have a pull request open: https://github.com/google/capirca/compare/master...jcatrouillet:capirca:master#diff-c11ad0bd48e47a1421118f82ccb0b4bdc5b943a36389e9c0f745e3f3a850d879R151-R160)

capirca/lib/nsxt.py
```diff
@@ -317,20 +356,22 @@ def _ParseFilterOptions(self, filter_options):
 
   def __str__(self):
     """Render the output of the nsxt policy."""
-    target_header = _RULE_JSON
-    target = []
-
+    # TODO: Handle when there are many policies
```
ivucica, line 359: Can you expand on what this would entail? Is it supposed to say "when there are too many policies" instead?

capirca/lib/nsxt.py
```diff
@@ -137,69 +189,55 @@ def __str__(self):
...
-      direction = 'IN_OUT'
-
-    tag = ''
-
-    ip_protocol = ''
-    if self.term.protocol:
-      ip_protocol = self.term.protocol
-
     rule = {
       "action": action,
```
ivucica, line 211: It would be good if we made the file consistent when it comes to string quoting. I suggest using single quotes. (Except for docstrings, of course.)

ivucica (approved):
I have a few small questions, but other than that I am happy. I will ask the relevant internal team to take a look at merging this.

capirca/lib/nsxt.py
jgangel: "notes" allows us to have more characters than "description"

jcatrouillet (commented):
Why are you asking for the section name?
It's asking for an additional parameter to the user, when the API does not support a patch on /api/v1/infra/domains/default/security-policies/
It would be more user friendly to not ask for the section name as the PATCH is allowed at the section level: /api/v1/infra/domains/default/security-policies/section_name

PR #342, opened 2023-04-18
Integrate my original code with the newly integrated NSX-T generator

jcatrouillet:
# Fix the wrong use of notes instead of description, notes are supposed to be used for rule lock/unlock
# Add comment keyword instead of notes (comments are translated now to description in NSX-T API)
# Check description length against its maximum
# Add check for maximum number of rules
# Add support for term expiration
# Add a more granular management of ip_protocol and not blindly set it to IPV4_IPV6
# Fix bug convert 0.0.0.0/0 to ANY
# Add support for source_excluded
# Add support for verbose keyword
# Add check for mixed rule (v4 and v6) with v4 only source and no non-ANY v4 destination and v6 only source and no non-ANY v6 destination
# Add check for maximum number of source and destination IPs
# Fix issue when str of Term return nothing
# Add support for numerical protocol
# Fix bug when there's a single port, don't convert it into a range

### DIFF
--- a/capirca/lib/nsxt.py
+++ b/capirca/lib/nsxt.py
@@ -13,13 +13,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+#
+# VERSION 1.01
+# Fix the wrong use of notes instead of description, notes are supposed to be used for rule lock/unlock
+# Add comment keyword instead of notes (comments are translated now to description in NSX-T API)
+# Check description length against its maximum
+# Add check for maximum number of rules
+# Add support for term expiration
+# Add a more granular management of ip_protocol and not blindly set it to IPV4_IPV6
+# Fix bug convert 0.0.0.0/0 to ANY
+# Add support for source_excluded
+# Add support for verbose keyword
+# Add check for mixed rule (v4 and v6) with v4 only source and no non-ANY v4 destination and v6 only source and no non-ANY v6 destination
+# Add check for maximum number of source and destination IPs
+# Fix issue when str of Term return nothing
+# Add support for numerical protocol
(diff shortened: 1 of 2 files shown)


ivucica:
Small heads up: I had to do an assortment of other fixes, so I did internal code changes way before seeing your pull request. Once it is submitted, I’ll try to do a rebase myself that you can squash into your pull request, since this is the second time the contribution is being preempted.

Apologies for this!

jcatrouillet:
Ack, please let me know when you're done.

capirca/lib/nsxt.py
```diff
@@ -13,13 +13,31 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+#
+# VERSION 1.01
+# Fix the wrong use of notes instead of description, notes are supposed to be used for rule lock/unlock
```
ivucica, line 18: This is an important explanation, this would be a clear reason to put the notes into description (possibly in addition to notes).

ivucica (commented):
I've added some notes for whoever does the followup here (whether myself or @jcatrouillet). These are not so much comment and feedback, of the type I would usually do in code review -- more notes on what was changed here, because so much of it was already done in previous commits.

The main thing to add would be the max length checks. Adding support for `verbose` requires some more consideration not to break existing terms/policies (including our existing internal integration).

Thank you for your contribution.

ivucica:
When approaching this, I might create a completely new commit/PR and appropriately use the [`Co-authored-by:` tag](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) in the commit description.

PR #356, opened 2024-09-17
workflows: Upgrade actions to latest versions.

ivucica:
This updates actions to:

* `actions/checkout@v4`
* `actions/setup-python@v5`
* `actions/setup-node@v4`
* `actions/upload-artifact@v4`
* `microsoft/setup-msbuild@v1.3.3`

This is because of the following error which we have been receiving in the CI:

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v2`. Learn more: https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/
```

This does not attempt to fix the pre-existing build issues caused by typing problems introduced in other commits.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v4
       - name: Log into registry
         run: |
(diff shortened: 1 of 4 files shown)


ivucica:
Note, this will be likely sent as an internal code change instead -- this is sent as a PR to try out external CI.

